PR #12, opened 2015-05-08
Added nicer usages of environments

dennie170:
With this new feature, it becomes super simple to use multiple environments.

First we create an .local.env or a .production.env, and secondly we create a .env file that only contains local or production.


rappasoft:
I'm confused. Wasn't this how 4.2 did it? Why do you need testing and production? There should just be one .env file that's gitignored, so which ever is in its place will be used. Isn't that the flow of things with L5?

dennie170:
Yes, this is basically how 4.2 works. I just don't like the concept of a singular .env file.
You can close the request..

PR #30, opened 2015-06-15
Set admin names on Admin LTE

blomdahldaniel:
As much as I love you and the repo Rappa, I would very much like to not have to change this every other time i clone your repo. 

What do you think?

### DIFF
--- a/resources/views/backend/includes/header.blade.php
+++ b/resources/views/backend/includes/header.blade.php
@@ -110,14 +110,14 @@
                       <!-- The user image in the navbar-->
                       <img src="{!!asset('img/backend/user2-160x160.jpg')!!}" class="user-image" alt="User Image"/>
                       <!-- hidden-xs hides the username on small devices so only the image appears. -->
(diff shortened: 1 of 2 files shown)


blomdahldaniel:
Sorry for not putting them in the same commit, it was such a small change so i did it here threw github and i didnt know how to make it propely..

rappasoft:
I think it should use the access wrapper, you can replace those with access()->user()->name and it should do the same. Then I can merge it in.

blomdahldaniel:
Okey so how and why the access wrappers? Just the syntax you would like in your reop or is there something other more useful in mind of that?

rappasoft:
The access wrapper provides access to the user control management system, plus it would keep it consistent. It is literally Auth::user()->name anyway. But with access you can do access()->can() access()->hasRole(), etc.

blomdahldaniel:
Aah, I see! Great, now I created a new pull request with the 2 file changes in the same branch.

PR #33, opened 2015-06-16
Added "fzaninotto/faker": "~1.4" to composer.json

blomdahldaniel:
This was missing since last commit where the database/factories/ModelFactory.php was added.


blomdahldaniel:
Can we think of any place else where the import of Faker could be missing something? It works for me right now as it is.

PR #35, opened 2015-06-22
Active menu state in admin-panel

blomdahldaniel:
Hi!
I don't know if this is the direction that you want to go. But i find it nice to have this [active menu plugin letrunghieu/active](https://github.com/letrunghieu/active).

This can be used on nav li items to set an class of active.
```<li class="{{ Active::pattern('admin/access/*') }}"><a href="{!!url('admin/access/users')!!}"><span>User      Management</span></a></li>```

In this case, the nav link to "User Management" stays active while you browse through roles and permission.

When i ran my composer update after my installation of letrunghieu/active my composer.lock file found some updates, dont know if you want to merge those in to so I made it in a separate commit in that case.

This is just a suggestion, I will keep on adding this myself every time i clone your repo so i figured that I might just ask you to merge it in.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,8 @@
         "stripe/stripe-php": "1.*",
         "twbs/bootstrap": "3.*",
         "twbs/bootstrap-sass": "3.*",
-        "fortawesome/font-awesome": "4.*"
+        "fortawesome/font-awesome": "4.*",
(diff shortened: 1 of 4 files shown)


rappasoft:
Looks good!

blomdahldaniel:
Glad you liked it. :)

PR #39, opened 2015-06-30
Create .env.example

blomdahldaniel:
Want to add this just for convenience, what do you think?
Makes the process of installing a bit smoother in my humble opinion.


rappasoft:
Sounds good!

PR #42, opened 2015-07-14
Google analytics id to env

blomdahldaniel:
I think that the repo does have a nice configuration factor in the proposed `.env.example`.
But I think that the Google Analytics ID also should be among them inside the .env

Maybe not because it has to do with the specific environment you are in but as I said i would like to have it there for the configuration factor. Or would you prefer to add it to the `config\app.php` file? 

Anyhow, I think it would feel right to add the GA ID to a config level.

I tried to follow what seems to be your convention in the .env and not write out default values inside the .env but where the value is called from. So in the `ga.blade.php` there is this line:
```{{ env( 'GOOGLE_ANALYTICS_ID', 'UA-XXXXX-X' )}}``` 

What do you think?


rappasoft:
I feel like this would be better off in the app.config or maybe a new analytics.config file, the reason being is because at least when I develop I don't have a different ID for different environments, and you can set the default to be the XXX ID.

But i'm open to either, maybe some one else can chime in with a thought?

blomdahldaniel:
Yes, I see what you mean. The environment file should be about the environment. But I do feel that the GA ID should be in a config (imo) but where to put it might be the question here as you say.

rappasoft:
I feel like if it were a stock thing it would be in something generic like config/analytics.php or something.

blomdahldaniel:
What do you mean by "stock thing"?

rappasoft:
Like if it came out of the box.

rappasoft:
I'm gonna make a config file for this one instead of the env variable. I'll add it today.

blomdahldaniel:
Ok I see. Yes i guess something like a config/analytics.php could be fitting.

I am planing to set up [spatie/laravel-analytics](https://github.com/spatie/laravel-analytics) and with that you should create a config/laravel-analytics.php file. That would then be a great place to put the Google Analytics ID. That might be something to look into maybe?

PR #43, opened 2015-07-14
Fix for bootstrap js source

blomdahldaniel:
I came across an issue where the modal black/faded background was placed above the modal-box. This was because of a diff from the current Bootstrap 3.3.5 css (from the vendor) and the 3.1.1 js that was fetched from cdn (in layout\master.blade.php). 

To ensure the same version of js and css, I added a .copy() in the `gulpfile.js`. which fetch bootstrap.min.js from
`vendor\twbs\dist\js\` to `public\js\vendor\bootstrap.min.js`.

I then changed the link to the js in both the frontend- and backend-master layout.

Donno if you want to keep the cdn but this way you wont have to worry about updating the cdn-link.
The choice to swap to cdn could then be the developer's decision..

What do you think?


blomdahldaniel:
@rappasoft sorry about my mess with the commits... Bah...

rappasoft:
I like this because we'll only have to update the composer.json file for bootstrap with each major release. Let me just test it then I will merge it.

rappasoft:
Unless there's another way to reference the latest bootstrap using a cdn? jQuery used to have a jquery.latest.js CDN, but I think they got rid of it, if bootstrap has one maybe we should use that, everyone seems to think the use of a CDN is better than loading your own.

blomdahldaniel:
Yes I thought that that would keep your back fre from future updates at the cost of cdn caching.
I wanna give credit to @WictorJohansson as well who was finding the bug issue with me.

blomdahldaniel:
Aha, yeah that would also do the trick I guess. Yeah I can see way cdn would be preferable.

blomdahldaniel:
But I can at the same time totally see way you should not always fetch the latest cdn without control since that would bring a risk of compatibility breaks. But this solution (with fetching from the vendor), if you update, you update both css and js in that case. And you would also do that in development which states that you will be looking out for potential breaking of the code caused by the update?

rappasoft:
I suppose you should be doing your composer updates and checking for issues on a development server first and not on production anyway, so even if there is a problem you should find it first.

 I don't want to go overboard but i'm almost imagining a config/bootstrap.php file where you can choose cdn true/false and the view renders which ever chosen. But then again I don't want this to be one of the overly bloated projects with a million config options that's hard to understand out of the box.

I'm torn.

blomdahldaniel:
Hm... I can feel with you man. It is the same thing about the Google Analytics ID thing. It could go in the .env, even better in config/app.php but it doesnt quite feel right.

Same thing here.
But I actually think this merge would at least eliminate the upcoming issues where the css and the js is incompatible. Cause we had quit a hassle with it before we saw what was wrong...

So I favour some quick change of this, either merge this, or update the version in your cdn link (to 3.3.5) and then keep thinking about how to go with it in the future.
I guess one way would also be to concatenate the main.js and the vendor/twbs js?

rappasoft:
Ok that's merged, and I just pushed an analytics config file.

I didn't merge the bootstrap js with the rest, reason being it still allows for the 'before-scripts-end' section which allows you to put assets in between the bootstrap js and the main js, if need be.

blomdahldaniel:
Great! :+1:

PR #55, opened 2015-08-11
Change .less for backend

TwiSka:
Modification for backend css with less


rappasoft:
Fixed it in the master copy because has to tweak other things, thanks.

PR #79, opened 2015-09-14
Changing the AuthController to send the links of active social networ…

andersondeoliveiramachado:


rappasoft:
Actually, I think it's fine for now. I'm going to go through the project in its entirety and refactor at some point when I redo the access library. Nice work.

andersondeoliveiramachado:
Thanks for accepting. I am new to laravel. Now I will try to make a simple crud and enter the menu. What do you think of the idea? maybe the tables countries and states frontend.layouts.master

rappasoft:
I'm not following. What exactly do you want to change?

andersondeoliveiramachado:
In fact would not change but add.

Two tables in the database: countries and states

And put a menu on the backend.

entries
   countries
   states

Can use the package:

https://github.com/mitulgolakiya/laravel-api-generator

rappasoft:
That may be a little overkill for this project. Those are just examples to extend the form helpers since they were removed in laravel 5.0.

Issue #98, opened 2015-09-22
.form-control:not(select) also excludes input[type=checkbox]


SchizoDuckie:
I just noticed that checkboxes generated with bootstrap forms don't render. 

The cause of this is 

```
form-control:not(select) {
     -webkit-appearance: none;
    -moz-appearance: none;
    appearance: none;
}```

By the way, this has to be the most awesome laravel boilerplate / admin out there. great work!

rappasoft:
Did you get my comment on the pull request?

PR #99, opened 2015-09-22
update main.css to fix checkbox exclusion for bootstrap form controls.

SchizoDuckie:
re #98


rappasoft:
The css files are rendered from either the sass or less files, so editing a css file will be immediately overwritten the next time the user ran gulp. Have to figure out which sass/less file it's in.

PR #121, opened 2015-10-05
Remove unnecessary use keywords.

lancepioch:
$router is no longer used beyond these definitions.


rappasoft:
But it's needed to be accessed in the closure from the other included route files. If you run what you have the application shouldn't work.

lancepioch:
You're right.

PR #123, opened 2015-10-06
Added carbon

jarod51:
Carbon is a so essential DateTime manipulation library :) https://github.com/briannesbitt/Carbon


rappasoft:
Doesn't carbon come with the default installation of laravel? If you reference \Carbon\Carbon::now() you have it, the created/updated timestamps are default carbon. And you have the protected $dates property on models to specify more.

jarod51:
You are right ! Sorry :) Forget my PR

PR #161, opened 2015-11-24
Localization to zh-CN

Hotren3rd:
Localization to zh-CN(本地化简体中文支持)


rappasoft:
Confused at what this is? There's no files changed.

Issue #141, opened 2015-10-29
Banning and deactivating a socialite user


brentkozjak:
Hi,

Awesome work on this boilerplate!  I've noticed a permission issue for users that are created through socialite.  When I log in as an administrator and ban or deactivate a social user, they are still able to login.  If I ban a 'local' user, I get the error "Your account is currently deactivated." which is obviously expected.  Has anyone run into this issue?

Cheers!
Brent

ChristianGiupponi:
Same for me

PR #165, opened 2015-11-26
Prevent user login with social account if banned or deactivate

ChristianGiupponi:
If user has been banned or deactivated by an Admin he can't log in with the social account.
This should fix: https://github.com/rappasoft/laravel-5-boilerplate/issues/141

I have also added a small fix in  postEmail method.  
In this way we can check if the email input is filled and the value is a valid email, everything before execute any query.

### DIFF
--- a/app/Http/Controllers/Frontend/Auth/PasswordController.php
+++ b/app/Http/Controllers/Frontend/Auth/PasswordController.php
@@ -52,6 +52,9 @@ class PasswordController extends Controller {
      */
 	public function postEmail(Request $request)
 	{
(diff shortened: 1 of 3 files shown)


rappasoft:
Wait, if the user is banned you want them to still be able to log in with a social account?

ChristianGiupponi:
No, don't want it, right now if user is banned can access with the social login.  
This fix should check if the account is banned and stop them to log in.

How to reproduce the bug:

1. Register with a social network.
2. Using the admin account ban the user.
3. Try to log in with the previous user and the previous social account. It can login.

PR #166, opened 2015-11-30
Fixed MAIL_ENCRYPTION in mail config. allow use it in .env file

TfmGroup:
mail encryption needs to be able to be set from .env file ( found the problem when i tried to use gmail account to send email ).


rappasoft:
Merged this, Don't know why it's still here so gonna close.

PR #172, opened 2015-12-02
added scaffolding generator from alnutile/l5scaffold

shiruken1:
This thing ROCKS for RAD!

https://github.com/alnutile/l5scaffold

### DIFF
--- a/app/Http/routes.php
+++ b/app/Http/routes.php
@@ -1,40 +1,6 @@
 <?php
 
 /**
- * Switch between the included languages
+ * Set Up Additional Routes Here
+ * Check RouteServiceProvider for defaults
  */
-$router->group(['namespace' => 'Language'], function () use ($router)
-{
-	require(__DIR__ . "/Routes/Language/Lang.php");
-});
-
-/**
- * Frontend Routes
- * Namespaces indicate folder structure
- */
-$router->group(['namespace' => 'Frontend'], function () use ($router)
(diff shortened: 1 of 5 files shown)


rappasoft:
What are all of the changed files doing?

shiruken1:
You can ignore this. The scaffolding system works, but you've put in more work (awesome job!) than it can automatically handle for you. I'll see about forking it and implementing it for your boilerplate in the future.

Cheers!

TfmGroup:
scaffolding system is not aware of backend / frontend structure. It generates the database migration / controller / blades but after that a lot of editing will be needed. took me an hour or so to get from:
hp artisan make:scaffold DigitalSign --schema="id:increments, name:string:default(''),
uid:string,
is_active:boolean:default(FALSE),
assigned_to:boolean:default(FALSE),
last_seen:timestamp,
client_id:timestamp,
location:string,
gps_long:float:default(0),
gps_lat:float:default(0),
timestamps "

to get it working in admin dashboard. It can be modified to fit in boilerplate but it's quite some work to be done. But i completely agree that can be a time saver if is implemented the right way.

shiruken1:
It shouldn't be too hard, either, just time consuming... and right now I'm trying to get this thing done and over with. The stub system it uses is pretty simple-looking! Hopefully I'll get to take it one once I'm done with the current project

PR #180, opened 2015-12-11
PSR-2 Standards

helmerdavila:


rappasoft:
All you did was follow PSR-2 coding standards you didn't physically change any functionality right?

rappasoft:
The only thing I dislike, is that I prefer my use statements to be in length order, I believe it is how Laravel does it. It is not stated in PSR-2 that they have to be alphabetized. Anyone else?

Even the sample has them in length order: http://www.php-fig.org/psr/psr-2/#3-namespace-and-use-declarations

helmerdavila:
Yes, only PSR-2 changes, nothing in functionality. But if you prefer, i can fix that (use statements).

rappasoft:
Yeah I just think the use statements looks so much nicer in length order. Other than that I don't see a reason I can't merge it after that.

helmerdavila:
Well. I added 2 commits more. Everything is right?

rappasoft:
I'm getting a syntax error on this line: https://github.com/helmerdavila/laravel-5-boilerplate/commit/dbed71895f121dee74839c32a0099f0693c0c2af#diff-e3c1bcc53ca2a6d4a736acd34862ec94R17

rappasoft:
I can fix it if you don't have time.

helmerdavila:
Fixed. Committed with --amend

rappasoft:
Merged thank you!

PR #209, opened 2016-01-02
French language pack

error500:


limatheus:
After this pull request other languages  are removed.

PR #214, opened 2016-01-03
Update .gitignore

ninjaparade:
add .DS_Store to ignore.


rappasoft:
This one I will manually do because it is in there but has an additional slash for some reason. Thank you.

Issue #230, opened 2016-01-07
Add pluralization to translations


blomdahldaniel:
Since it is pretty easy to add pluralization to translations I see no use in having translations like this:
````
'total' => 'permissions(s) total',
````

With the pipe function (see [/docs/5.2/localization#pluralization](https://laravel.com/docs/5.2/localization#pluralization) )

````
'total' => 'permission total|permissions total ',
````

In the view:
````
// From:
{{ $roles->total() }} {{ trans('labels.backend.access.roles.table.total') }}

// To:
{{ $roles->total() }} {{ trans_choice('labels.backend.access.roles.table.total', $roles->total() ) }}
````

I could do this PR if you think it is an improvement that should be done.
I guess this would be the same for 
* User(s)
* Permission(s)
* Role(s)

I can take a look and search for more pluralizations and fix them as well. **It is easy now before there comes in more translations.**

rappasoft:
Yeah i'd say let me pull the french changes and your swedish pack and after that's all pushed if you wouldn't mind doing all three that would be great.

rappasoft:
Merged!

PR #232, opened 2016-01-07
locale pluralization for number of rows in access tables

blomdahldaniel:
Pluralizations for the access tables. 
**Note** 2 of the Swedish translations have the same word independent of singular or plural. Just wanna inform that it is not a duplication.

### Warning for fr-FR though
This PR will not break anything. But, one possibly major thing is that the pluralization trans_choice() does not seem to work at all for the fr-FR version. Soo strange! So as of now, the singular sting is shown all the time for French and only French. I checked the Swedish words that doesn't change normally and that worked. And the English version works just as expected. So its something really fishy about the fr-FR translation here.. I dont understand!

### DIFF
--- a/resources/lang/en/labels.php
+++ b/resources/lang/en/labels.php
@@ -58,7 +58,7 @@ return [
                     'permission' => 'Permission',
                     'roles' => 'Roles',
                     'system' => 'System',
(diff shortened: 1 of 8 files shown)


blomdahldaniel:
solves #230

Issue #211, opened 2016-01-03
Locale on Carbon and date formats


error500:
Hello
I don't know if the Carbon date for humans should works with current locale.... Is it a problem on my installation ?
setlocale(LC_TIME, 'fr_FR');

Arrakeen:
You could use https://github.com/LaravelRUS/localized-carbon
However, the French translation is probably incomplete.

blomdahldaniel:
Hm.. I am not sure how this has been solved before. One way to set the carbon locale is to put 
````
//will give swedish diffForHumans and time
    setlocale(LC_TIME, "sv_SE");
    date_default_timezone_set('CET');
    Carbon::setLocale('sv');

````
Inside the `Providers\AppServiceProvider`.

Not sure how it should be solved in this repo though

petergit67:
Is this correct:

    public function boot()
    {
        setlocale(LC_TIME, "sv_SE");
        date_default_timezone_set('CET');
        Carbon::setLocale('sv');
    }

Works fine anyway.

blomdahldaniel:
Yes, inside the boot method. I am sure @rappasoft will figure out a nice way to deal with this with his current language system.

petergit67:
@blomdahldaniel are you working on an Swedish translation?

blomdahldaniel:
@petergit67 ja just nu faktiskt, har kommit en liten bit iaf. Jag taggar dig i PR så kan du syna och ge förslag?

rappasoft:
Jag förstod bara detta på grund av google :)

petergit67:
Sorry we will not write in Swedish here. Daniel, could i have your mail so i can communicate in Swedish directly with you?

blomdahldaniel:
Excuse me guys! Thought that it could be fun to write some swedish on github.. ;) Keeps rappa on his toes.. ;)

@petergit67 mail@blomdahldaniel.se

@rappasoft I have started to think that the `protected $languages` array needs to be different. I think that many different languages has multiple (often 2) names for the language. As in this example above, for us to get swedish in both php in general and in Carbon we need both the strings "sv_SE" and "sv"

So the `$languages` array in `\LocaleMiddleware` maybe has to bee different?
Something like this:
[ carbon-specific, php-language-code ]

Here is a list of Carbon language codes:
https://github.com/briannesbitt/Carbon/tree/master/src/Carbon/Lang

Here is a list of php/general locale codes
[...]

PR #233, opened 2016-01-08
fix for carbon and php localization regarding dates

blomdahldaniel:
This solves #211
Enables localized `diffForHumans()` and `Carbon::now()->formatLocalized()`

# Report on locales
I feel that there is a lot to talk about when it comes to this. Therefore I have tried to gather some info on the subject and try to explain why this is important and how it works.
I have also done some research on language standards for you to take a stand on for the future.

#### What does setLocale() and Carbon::setLocale() do?
**I took french as the example:**
1. Carbon diffForHumans() - **il y a 10 heures** (10 hours ago)
2. Carbon::now()->formatLocalized('%e %B') - **8 janvier** (8 January)

**Solution**
1. `Carbon::setLocale('fr');`
2. `setLocale(LC_TIME, 'fr_FR');` (see [Carbon Docs](http://carbon.nesbot.com/docs/#api-localization))

### 1. Carbon
Carbon uses the ISO-639 standard (**ISO-639-1** to be exact).
[List of ISO-639 locale codes](http://www.loc.gov/standards/iso639-2/php/code_list.php)

**Examples**
* `en` - English
* `fr` - French
* `sv` - Swedish

### 2. PHP
PHP uses the GNU C Library formatting for locale codes. It combines languages and regions to enable more specific libraries. It also uses the **ISO-639-1** for both the *language* and the *region* codes. Language is separated with `_` from region and the region is spelled with capitals.
[List of GNU locale codes](http://lh.2xlibre.net/locales/)

**Examples** *language_REGION*
* `en_US` - English - United States
* `fr_FR` - French - France
* `fr_CH` - French - Switzerland
* `sv_SE` - Swedish - Sweden

--------------------------------------------------------------------------------
# Consideration
1. What standard should this boilerplate follow regarding naming of folders and key name in the `$languages` array?

### 1. Diversity
To enable more specific translations the combined *language* + *REGION* system is great.
However, now that we have to follow PHP GNU C Library formatting in one case, I think we should use that standard also in the folder names when necessary. *When necessary* I mean with languages that possibly could have multiple regions that wants an translation for example Portuguese, Spanish and French.

##### Pros And Cons
**Pros**
1. Great possibilities to grow and have very specific languages
**Cons**
1. Honestly, possibly a whole lot of poorly updated language packages
2. one pretty ugly array..
```
protected $languages = [
    'en' => ['en', 'en_US'],
    'es_ES' => ['es', 'es_ES'], (spanish - Spain)
    'es_AR' => ['es', 'es_AR'], (spanish - Argentina)
    'fr_FR' => ['fr', 'fr_FR'],
    'pt_BR' => ['pt', 'pt_BR'], (Portuguese - Brazil)
    'pt_PT' => ['pt', 'pt_PT'], (Portuguese - Portugal)
    'sv' => ['sv', 'sv_SE'],
];
```

### 2. Lets be real
Say screw it and only allow one translation per language. Because sure, this repo is getting bigger but it doesn't need to be that language specific? Let's not get things overly complicated?

##### Pros And Cons
**Pros**
1. More simple and slick folder structures
2. More likely to be updated better while less languages
3. A darn fine and easy to manage `$languages` array
**Cons**
1. lost posibility of specific languages
2. Small refactor needs to be done:
* Rename all occurrences of `fr-FR``
. 
* Refactor the `LocaleMiddleware` a litle
```
protected $languages = [
    'en' => 'en_US',
    'es' => es_ES',
    'fr' => 'fr_FR',
    'pt' => 'pt_PT',
    'sv' => 'sv_SE',
];
```
--------------------------------------------------------------------------------

What do you say?
I would say option 2 and then good night to you good sir.

### DIFF
--- a/app/Http/Middleware/LocaleMiddleware.php
+++ b/app/Http/Middleware/LocaleMiddleware.php
@@ -15,9 +15,9 @@ class LocaleMiddleware
      * @var array
      */
     protected $languages = [
-        'en' => ['en', 'en_US'],
+        'en'    => ['en', 'en_US'],
         'fr-FR' => ['fr', 'fr_FR'],
-        'sv' => ['sv', 'sv_SE'],
+        'sv'    => ['sv', 'sv_SE'],
     ];
 
     /**
--- a/app/Http/Middleware/LocaleMiddleware.php
+++ b/app/Http/Middleware/LocaleMiddleware.php
@@ -3,6 +3,7 @@
 namespace App\Http\Middleware;
 
 use Closure;
+use Carbon\Carbon;
 
 /**
  * Class LocaleMiddleware
@@ -13,7 +14,11 @@ class LocaleMiddleware
     /**
      * @var array
      */
-    protected $languages = ['en', 'fr-FR', 'sv'];
+    protected $languages = [
+        'en' => ['en', 'en_US'],
+        'fr-FR' => ['fr', 'fr_FR'],
+        'sv' => ['sv', 'sv_SE'],
+    ];
 
     /**
      * Handle an incoming request.
@@ -22,12 +27,16 @@ class LocaleMiddleware
      * @param  \Closure                 $next
      * @return mixed
      */
-    public function handle($request, Closure $next)
-    {
-        if (session()->has('locale') && in_array(session()->get('locale'), $this->languages)) {
-            app()->setLocale(session()->get('locale'));
-        }
+    public function handle($request, Closure $next){
+        if (session()->has('locale') && array_key_exists(session()->get('locale'), $this->languages)) {
+                app()->setLocale(session()->get('locale'));
 
+                // setLocale for php. Enables ->formatLocalized() with localized values for dates
+                setLocale(LC_TIME, $this->languages[ session()->get('locale') ][1] );
+
+                // setLocale to use Carbon source locales. Enables diffForHumans() localized
+                Carbon::setLocale( $this->languages[ session()->get('locale') ][0] );
+            }
         return $next($request);
     }
 }


rappasoft:
I applaud your research and will go through it throughly probably tomorrow and test it on my local version. I'm all for what's cleaner but also what will prevent me from having to refactor in the future. I want it to be easy for people to contribute language packs, which means option 2 is probably the cleanest.

The only thing I'm worried about is for example when having 1 pack for all, people from (spanish - Spain) making a PR to change it because it's wrong for them because its set for (spanish - Argentina). Then there will be PR's back and forth for the same language pack because theirs different dialects. 

If that ends up being a problem it would mean theres a shit ton of people using this. I honestly have no idea how many people currently use it.

I'm really glad you had the time to research, because i've been researching the timezone issues.

There are all things that we take for granted when using online services because they always seem to have all this stuff, and I never bothered to see if there was a norm or how they did it.

Eventually I hope to have this be such a solid complete solution as a starting point for not only myself, but everyone. A good thousand hour head start at this point.

blomdahldaniel:
I almost think this boils down to a decision from you. What signals you want this Boilerplate to send. My hones opinion is that I do not think that it should/could be that big of a problem if we only us `fr` and `es`. 

But if you want to keep it all 100% flexible I cant blame you. But for the sake of consistency you might apply the *language + REGION* system for all the languages directories (just like the latest issue talked about). And because of the PHP GNU C library you could name the  folders with the same separator,  `_` ?
```
resources/lang/
    en_US/
    sv_SE/
    fr_FR/
    es_ES/
// etc....
```

It is a bit funny how the entire boilerplate talks about the locale stuff now, namings, language codes and stuff. Haha! :D kind of cool to see so much interest. :)

Let me know when you have taken a look at this and comes to conclusion.

blomdahldaniel:
## Big question though
Setting the language in the `session()` is not a great way of doing it in my honest opinion. It doesn't take long before the session is lost and the language defaults back to english. Maybe about 6 hours or something is my experience with it now. Is it possible to store cookies on the user end without a cookie service/driver?

error500:
I would say cookies for guest users and database for identified users

rappasoft:
Been swamped with work but I promise i'll get to these soon.

blomdahldaniel:
@rappasoft no stress bro!

blomdahldaniel:
This was a quick and dirty fix for this issue. I think we hav come up with a great solution now. So I will close this out.

PR #239, opened 2016-01-13
Localisation comment instructions

blomdahldaniel:
When i was adding my own language. I felt that the process was a bit "scary" or at least I did not feel very comfortable it at first. I was double checking a lot that I was in the right file etc..

These comments are supposed serve as a comforting guide in the process of adding a new language.


blomdahldaniel:
Feel free to ask me to change stuff/change it yourself if you have any documentation you would like to "rappasoftify" :p

PR #252, opened 2016-01-21
Remove log-viewer from resources/views/vendor

### LINKED ISSUE #251: Remove log-viewer from resources/views/vendor
This view should not here, it should depend on log-view packages because this will always change when developer run `php artisan vendor:publish` which publish other packages to project.

toancong:
close #251


toancong:
I close because I wrong about this.

PR #260, opened 2016-01-25
Add german translation

PReimers:
I've added the german translation and fixed some of the TODOs in other translations

PR #261, opened 2016-01-26
Translate menus (german)

PReimers:
Fixed the issue you mentioned in #260


PReimers:
That was a fast merge.. :)

rappasoft:
Figured while I have the time I might as well get it all in!

PR #273, opened 2016-02-03
tables timestamp fields : removed unecessary default values

error500:
This defaults value is not available in postgres
$table->timestamp('updated_at')->default('0000-00-00 00:00');
And 
$table->timestamp('updated_at')
will anyway push the  default value through eloquent save function.
I remove also the default on field created_at for homogeneity reasons

### DIFF
--- a/app/Http/Controllers/Backend/Access/User/UserController.php
+++ b/app/Http/Controllers/Backend/Access/User/UserController.php
@@ -5,6 +5,7 @@ namespace App\Http\Controllers\Backend\Access\User;
 use App\Http\Controllers\Controller;
 use App\Repositories\Backend\User\UserContract;
 use App\Repositories\Backend\Role\RoleRepositoryContract;
+use App\Http\Requests\Backend\SearchRequest;
 use App\Http\Requests\Backend\Access\User\CreateUserRequest;
 use App\Http\Requests\Backend\Access\User\StoreUserRequest;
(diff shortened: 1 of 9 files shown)


blomdahldaniel:
Then, if we are safe to go back to the pure laravel setting for this I think it is better looking to use the laravel `$table->timestamps()`

So I suggest that we change this in all migrations as well:
```diff
- $table->timestamp('created_at');
- $table->timestamp('updated_at');
+ $table->timestamps();
```

rappasoft:
There was an issue with the default timestamps() method, that's why it was changed. I just don't remember exactly what it was.

rappasoft:
I believe the issue was both columns were getting an on update current timestamp attribute for no reason.

blomdahldaniel:
Hm.. @error500 the last commt a9a5599 maybe should be in its own PR and not in this one?
It looks interesting though

error500:
@blomdahldaniel yes, i'm a github & pull request newbie...

blomdahldaniel:
No worries. Then you fit in with the rest of us. :)

error500:
All my commits on branch master will go on the same PR ? is that the reason i should create 1 branch for each PR purpose ?

blomdahldaniel:
Yep exactly! :)
One branch for every pull request. Preferably named in a similar way as the PR.

rappasoft:
Closing this since there are new separate requests.

PR #275, opened 2016-02-04
Update header.blade.php

kovacbb:
Real user "created at" date


blomdahldaniel:
### This is a good idea.
Ther is one catch though. The date format. Some would say that the format should be `format(YYYY/mm/dd)` and I am one of those guys.. :p

I know that @rappasoft is working with a solution to this. So the selected language should set what format that should be used. **Because the format of the date is bound to specific languages.**

I am not sure but I think that we might wait with this one until the locale date format stuff is done. And then put it on our to do list. :smile:

rappasoft:
Yeah I think we will wait otherwise it will be one more thing to forget. But i'll add it to the list.

PR #280, opened 2016-02-06
backoffice search user enable

error500:
HI.
I made the search field functionnal in backoffice users list.
It looks like a test for me. Don't hesitate to reject if it's not correctly wrotten or not in road map.


rappasoft:
I appreciate this request, but I think it may go a little beyond the bounds of this boilerplate. I think it's up to each person on how they would want to search and make their own API, instead of being stuck to one either I or one of us make. But good job regardless.

error500:
Ok Anthony, I understand your point of vue.
And what about sortable columns ?

Issue #276, opened 2016-02-04
error when db:seed


hopewise:
```
[Illuminate\Database\QueryException]                                                                                                                                                                                                           
  SQLSTATE[23503]: Foreign key violation: 7 ERROR:  insert or update on table "permission_dependencies" violates foreign key constraint "permission_dependencies_permission_id_foreign"                                                          
  DETAIL:  Key (permission_id)=(22) is not present in table "permissions". (SQL: insert into "permission_dependencies" ("permission_id", "dependency_id", "created_at", "updated_at") values (22, 1, 2016-02-04 11:30:09, 2016-02-04 11:30:09)) 
```

I found that in file: `database/seeds/Access/PermissionDependencyTableSeeder.php` line 41

`for ($i = 3; $i <= 23; $i++) {`

where the maximum id number for table permissions where 21

daniel-farina:
Did you manually remove rows from the database? It seems that you are trying to add a record that depend on permission_id)=(22) in your permissions table.

So the permission_id = 22 is missing in your permissions table.

You can always delete all the database migrate and seed again. 

Dan

hopewise:
I didn't remove any rows, its that when db seeded, rows are missing, so it
got out of range in the loop .. I don't know why the loop is for 23? and
what are those two missing records at permissions table are?

On Thu, Feb 4, 2016 at 4:03 PM, Dan <notifications@github.com> wrote:

> Did you manually remove rows from the database? It seems that you are
> trying to add a record that depend on permission_id)=(22) in your
> permissions table.
>
> So the permission_id = 22 is missing in your permissions table.
>
> You can always delete all the database migrate and seed again.
>
> Dan
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/rappasoft/laravel-5-boilerplate/issues/276#issuecomment-179853106>
> .
>



-- 
--
Kind Regards,
---------------------------------------------
Samir Sabri
Software Architect& Developer
www.dcaclab.com
Jordan-Middle East

blomdahldaniel:
Do you have the ability to run a refresh on the DB and then seed again?
```bash
php artisan migrate:refresh
php artisan db:seed
```

hopewise:
I already did so, to avoid the error, I had to alter the loop to count for
21 instead of 23



On Fri, Feb 5, 2016 at 1:38 AM, Daniel Blomdahl <notifications@github.com>
wrote:

[...]

Issue #250, opened 2016-01-21
Problem DB_CONNECTION=pgsql


devillom:
After migrate tables i have error date/time and fix it remove columns default values in migration files
After db:seed i have new problem 
```
[Illuminate\Database\QueryException]
  SQLSTATE[23503]: Foreign key violation: 7 ERROR:  insert or update on table "permission
  _dependencies" violates foreign key constraint "permission_dependencies_permission_id_f
  oreign"
  DETAIL:  Key (permission_id)=(22) is not present in table "permissions". (SQL: insert i
  nto "permission_dependencies" ("permission_id", "dependency_id", "created_at", "updated
  _at") values (22, 1, 2016-01-21 07:48:15, 2016-01-21 07:48:15))



  [PDOException]
  SQLSTATE[23503]: Foreign key violation: 7 ERROR:  insert or update on table "permission
  _dependencies" violates foreign key constraint "permission_dependencies_permission_id_f
  oreign"
  DETAIL:  Key (permission_id)=(22) is not present in table "permissions".
```
How to fix that?

blomdahldaniel:
Are your tables empty before seeding?

devillom:
Yes, I think this problem laravel core.

blomdahldaniel:
So it has to do with laravel core working with pgsql ?

devillom:
laravel has some bugs pgsql

rappasoft:
I don't think theres any good way for me to deal with the quirks on each connection type :(

error500:
Does it mean that the way is to turn off contraints as it is done in mysql ?
```
BEGIN
SET CONSTRAINTS ALL  DEFERRED 
... run seeders
COMMIT

blomdahldaniel:
I am going to close this issue as of now, it is out of hands for us to solve. Hope you found your way of solving this.

error500:
I just remove the defaults value. It seems to be unecessary to set this as eloquent push the current timestamp while create action on both fields
PR if you wish to make this change

PR #284, opened 2016-02-09
Remove hardcoded value (21) in PermissionDependencyTableSeeder.php

kdhana:
Handles future issues related to #276 and #250.


blomdahldaniel:
Hm.. I think this was fixed yesterday with #281 or do you think that this PR does something else than that.

error500:
@kdhana  did an elegant solution

kdhana:
@blomdahldaniel
This is a minor improvement and could be easily ignored.
Hardcoding values should always be avoided (even in such trivial scenarios).

blomdahldaniel:
Great job! @kdhana 
I am not so familiar with this issue so I will let @rappasoft take a look at it.

rappasoft:
I like it.

PR #294, opened 2016-02-15
Revert "defaulttimestamp"

FuzzyHobbit:
This reverts commit 82cf79650656608f3dc518e4090a27c98341b993.

MySQL before 5.7 does not accept more than one default CURRENT_TIMESTAMP per table. Repositories for popular distributions are still using 5.5 and 5.6.


rappasoft:
The problem is this:

- The default timestamps() seems to have 'on update current timestamp' for both created and updated, that I can't figure out.
- If I use 0000-00-00 00:00 as a string, it breaks on one database engine.
- If I use more than one CURRENT_TIMESTAMP, it doesn't work on MySQL < 5.7

We need a more elegant solution.

timgws:
@rappasoft please see #300, which I believe is a fix for this issue.

rappasoft:
Resolved with https://github.com/rappasoft/laravel-5-boilerplate/pull/300

PR #297, opened 2016-02-17
Fixed models __construct method

soulshockers:
Models parent constructor should be called in case child constructor is defined.

### DIFF
--- a/app/Models/Access/Permission/Permission.php
+++ b/app/Models/Access/Permission/Permission.php
@@ -29,10 +29,11 @@ class Permission extends Model
     protected $guarded = ['id'];
 
     /**
-     *
+     * {@inheritDoc}
      */
-    public function __construct()
+    public function __construct(array $attributes = [])
     {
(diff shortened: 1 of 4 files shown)


rappasoft:
Do you have an example as to why?

soulshockers:
/** @var \App\Models\Access\Role\Role $role */
$role->permissions()->create(['group_id' => 1, 'name' => 'test_name', 'display_name' => 'test_display_name']);
//or
Role::create([...attributes...]);
All attributes passed to create method will be lost.

rappasoft:
Are you sure? You should be able to use ```$role->permissions()->create``` without a problem?

soulshockers:
Of course I'm sure. 
method $role->permissions()->create is method of Illuminate\Database\Eloquent\Relations\BelongsToMany which call newInstance method of Illuminate\Database\Eloquent\Model which create instance of Permission model by passing $attributes array to our model __construct method. That's why parent constructor of every model should be called.

PR #300, opened 2016-02-19
Remove the default value on updated_at columns

timgws:
With MySQL 5.5, you should not have a default value of `CURRENT_TIMESTAMP` on more then one column at the same time. Also, a column added with no default

>  it is the same as specifying both `DEFAULT CURRENT_TIMESTAMP` and `ON UPDATE CURRENT_TIMESTAMP`.
>              (https://dev.mysql.com/doc/refman/5.5/en/timestamp-initialization.html)

This PR will remove the default current timestamps, and ensure that no default value is set on them.


timgws:
This will set CURRENT_TIMESTAMP as the default for created_at, and then `on update current timestamp` for the `updated_at`. If this is not acceptable, we can change the `->default()` for `updated_at` to be null.

PR #314, opened 2016-02-29
Move sensitive information to .env file

rdrubis:
Default application usernames and passwords should not be included in
version control. In case the .env file does not include the user
defaults, fallback values are provided as defaults.


DeeJaVu:
but it's just a seeder class to create users there's nothing sensitive

rdrubis:
Well at least for me, the default username and password are sensitive and should not be included in version control. This may also introduce security breach when users forget to change it after deploying in production.

rappasoft:
But they're dummy emails and passwords and are meant to be changed? Who's keeping admin@admin.com?

rdrubis:
No one. But that is not the point. Having that dummy email and passwords included in commits is not a good practice. The reason is similar to why Laravel stores database username and password in the .env file instead of having it on config/database.php.

rappasoft:
I think this is unnecessary but i'll let other people chime in. I personally don't want that information in my .env file because it's not really environment specific.

DeeJaVu:
@rappasoft i agree with you

rdrubis:
Alright, thanks anyway..

Issue #265, opened 2016-01-31
[Question] Social Logins Table


ninjaparade:
@rappasoft would you have any interest in storing more information in the social logins table?
I build a lot of apps that need signed requests which require the users token once authed.

It would require altering the `social_logins` table and updating the `findOrCreateSocial` method

If you're okay with it I'll send a PR.

rappasoft:
Does this work for any provider?

ninjaparade:
@rappasoft every provider would return a token. 

Socialite uses this in OAuth 2

https://github.com/laravel/socialite/blob/2.0/src/Two/User.php

and for OAuth 1

https://github.com/laravel/socialite/blob/2.0/src/One/User.php

so the token is available and returned for all providers.

rappasoft:
If you think it's something everyone would need and can take advantage of, as well as it won't change anything for the frontend user, then yeah make a PR.

blomdahldaniel:
@ninjaparade are you working on a PR for this? Or should i close this issue for now?

ninjaparade:
@blomdahldaniel I am going to submit one, sorry I am bogged down with a couple projects.
I have a couple with this implemented so it won't be much effort to submit this.

sorry its taken so long :dancer: :dancer:

blomdahldaniel:
No problem dude, we all got things on our plate outside of this project. No worries, I was just wondering so that we can try to keep the issue-list clean. :)

rappasoft:
Merged.

ninjaparade:
:dancer: :dancer: :dancer: :heart:

PR #315, opened 2016-03-05
Feature/social logins

ninjaparade:
this PR adds the feature discussed in issue #265 

This PR adds 2 columns to the `social_logins` table 
- adds token (to be used for authenticated calls beyond authentication)
- adds social avatar, to display users avatar from their social connection.

classes changed/updated.

`App\Repositories\Frontend\UserEloquentUserRepository@findOrCreateSocial`

- new `SocialLogin` now accepts `token` and `avatar`
- returning users `SocialLogin` now updates `token` and `avatar`
- `SocialLogin` with `NULL` email address now defaults to `"{$provider_id}@{$provider}.com"`

this has been tested with Twitter, Facebook and Github Auth.

### DIFF
--- a/app/Repositories/Frontend/User/EloquentUserRepository.php
+++ b/app/Repositories/Frontend/User/EloquentUserRepository.php
@@ -147,7 +147,17 @@ class EloquentUserRepository implements UserContract
             $user->providers()->save(new SocialLogin([
                 'provider'    => $provider,
                 'provider_id' => $data->id,
+                'token'       => $data->token,
+                'avatar'      => $data->avatar,
             ]));
+        }else{
+             /**
+             * Update the users information, token and avatar can be updated.
+             */
+            $user->providers()->update([
+                'token'       => $data->token,
+                'avatar'      => $data->avatar,
+            ]);
         }
 
         /**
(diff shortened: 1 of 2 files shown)


ninjaparade:
OH also :dancer: :dancer: :dancer:

blomdahldaniel:
this looks neat! :smile:

rappasoft:
Seems simple, if you think it will be of use. What would be a specific instance this would be needed?

Also:

```'email' => $data->email ? : "{$data->id}@{$provider}.com",```

Does every provider provide that type of email address? Just curious as I don't know much about it.

ninjaparade:
@rappasoft not every provider sends back an email, for example twitter doesn't so in this case, it creates an email that would be unique in the users table based on the users provider and provider id.

This also solves the issue of a some MySQL versions (I've encountered this) where more than 1 `null` value would be considered a not unique.

also, the biggest usage of this, is once the user is authed, you can use the token to make subsequent calls.

rappasoft:
Looks good thanks!

blomdahldaniel:
I guess this issue is ready to be closed..?

PR #333, opened 2016-03-23
Feature/analytics fix

vpratfr:
While Google Analytics is not configured, do not output the corresponding JS.

You may only want to incorporate a single commit: 	1e35825


rappasoft:
Good idea, I will just implement separately instead of trying to pull it out of your commit history though.

Thanks!

Issue #336, opened 2016-03-25
Resend confirmation email reveal secret token


therezor:
I can confirm my email just using token from resend confirmation link

` public function resendConfirmationEmail($token) {
        return $this->sendConfirmationEmail($this->findByToken($token));
    }`

Should not be findByToken.  $token is secret and should be displayed in email only.

Suggest use user id,login or email in resend link

rappasoft:
Where else would you see it?

therezor:
Steps to reproduce (register with fake email):
1. Register as usual
2. Try to log in, you'll get message "Your account is not confirmed. Please click the confirmation link in your e-mail, or click here to resend the confirmation e-mail."
3. Copy link "http://test.dev/account/confirm/resend/1866048e9c87542268fab2be49e679ad"
4. Remove "resend" from it -> http://test.dev/account/confirm/1866048e9c87542268fab2be49e679ad
5. Go to this url (http://test.dev/account/confirm/resend/1866048e9c87542268fab2be49e679ad)
6. You've got confirmed fake email

rappasoft:
If you have time can you fix and make a PR? I just made a big update and have to go for the day.

therezor:
@rappasoft Yes, Created pull request

rappasoft:
https://github.com/rappasoft/laravel-5-boilerplate/commit/3cf794c56dc2b2d9575f5c44785e4fb42e8e8a8f

PR #337, opened 2016-03-25
Hide secret token in resend confirmation link

therezor:
Fixed security problem  described here https://github.com/rappasoft/laravel-5-boilerplate/issues/336


rappasoft:
Isn't having the user id in the route just as bad?

therezor:
@rappasoft I don't think that user_id is very private information, so we can use login or email instead

rappasoft:
I merged this, but for some reason it didn't register so i'm going to close it manually. Thanks for the help.

PR #344, opened 2016-03-30
Exclude node_modules from backup

sebastiandedeyne:
Excluded the `node_modules` folder from the backup.

Not in this PR, but I'd also opt to not exclude `storage_path()`, since user uploads tend to be stored there sometimes.

### DIFF
--- a/config/laravel-backup.php
+++ b/config/laravel-backup.php
@@ -28,6 +28,7 @@ return [
                  */
                 'exclude' => [
                     base_path('vendor'),
+                    base_path('node_modules'),
                     storage_path(),
                 ],
             ],


rappasoft:
How about we include the storage app directory, and exclude the rest?

sebastiandedeyne:
Yeah that sounds best, you could just explicitly include `storage_path` and leave `exclude` empty.

rappasoft:
This is what I did: https://github.com/rappasoft/laravel-5-boilerplate/blob/f5f6ba1644112a7856fb38099177bb5637cac145/config/laravel-backup.php

Issue #349, opened 2016-04-04
The user's resend confirmation email link is not working


tianissimo:
According to #343, The route is under 'admin' middleware so it doesn't work for new registered users.

blomdahldaniel:
Hm.. Can you make a PR that solves this?

rappasoft:
This was fixed.

tianissimo:
@rappasoft No, I don't think it was fixed. I can still reproduce the issue. Here is the steps after installing the latest version of laravel-5-boilerplate:

1. Register a new user and don't confirm the email yet.
2. Try to login with the newly registered user.
3. On the login page, the error message ` Your account is not confirmed. Please click the confirmation link in your e-mail, or click here to resend the confirmation e-mail.` will be shown
4. Click "click here" from the error message above.
5. You will never get the confirmation email from this resend link.

Again, the route name `admin.account.confirm.resend` used in the translation key `exceptions.frontend.auth.confirmation.resend` works for admin only, not for normal users.

PR #358, opened 2016-04-14
Fixed frontend resend confirmation email route.

tianissimo:
The fix for the issue #349


Arrakeen:
I'm not really sure this needed to be fixed? I didn't have any issue with this functionality so far.

rappasoft:
He is right that it's wrong, I just feel like I did something else with this not too long ago.

PR #366, opened 2016-04-27
Update messages.blade.php

ninjaparade:
I get errors from the `$errors` not being set.
this update ensures that its present before continuing.


rappasoft:
Merged this but it didn't trigger.

PR #395, opened 2016-06-17
fix social user with no email

zhangmx:


rappasoft:
Manually merging this into the next release.

Thanks.

PR #417, opened 2016-07-13
Fix being able to chain temp logins.

Colbydude:
You're able to chain "logging in as" other users, then when you do, when you opt to log back in as the admin user, you're stuck as someone who may potentially not be you.

My solution simply overwrites the temp_user_id in the session and logs in as that user if they are not the same as the currently mocked user, or the admin user.

Edit: My PHP CS Fixer also did some automated code styling.

### DIFF
--- a/app/Repositories/Backend/Access/User/EloquentUserRepository.php
+++ b/app/Repositories/Backend/Access/User/EloquentUserRepository.php
@@ -43,20 +43,22 @@ class EloquentUserRepository implements UserRepositoryContract
         $this->user = $user;
     }
 
-	/**
+    /**
      * @param int $status
      * @param bool $trashed
      * @return mixed
      */
-    public function getForDataTable($status = 1, $trashed = false) {
-		/**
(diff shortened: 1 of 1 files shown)


rappasoft:
The code styling is throwing me off, did you just edit the login/logout as methods?

Colbydude:
I added this code block to the `App\Repositories\Backend\Access\User\EloquentUserRepository` loginAs method.

```php
// Overwrite who we're logging in as, if we're already logged in as someone else.
if (session()->has('admin_user_id') && session()->has('temp_user_id')) {
    // Let's not try to login as ourselves.
    if (auth()->id() == $user->id || session()->get('admin_user_id') == $user->id) {
        throw new GeneralException('Do not try to login as yourself.');
    }

    // Overwrite temp user ID.
    session(['temp_user_id' => $user->id]);

    // Login.
    access()->loginUsingId($user->id);

    // Redirect.
    return redirect()->route("frontend.index");
}
```

rappasoft:
Good catch.

Colbydude:
No problem. Glad to help out. 😃

PR #418, opened 2016-07-14
Change redirect of users based on their permissions.

danswiser:
To me it seems more convenient that the users with the 'view-backend' permission should be redirected to the backend dashboard while all other users continue to be sent to the frontend dashboard.


rappasoft:
Why the backend user page on the second one and not the dashboard again?

danswiser:
I'm honestly still new to git and didn't intend on including the second commit.

However, my thought there was taking the user back to the last page they would have seen before they chose to switch users.

rappasoft:
I do like both of your commits, but something is going wrong on my end and I have to merge them in manually. Thanks for the contribution.

PR #420, opened 2016-07-14
Added reCAPTCHA to Login Screen

poseso:
I added reCAPTCHA to the login screen, since i don't only use it in registration form, so this way we have it on the login and register form, activated via env

### DIFF
--- a/app/Http/Requests/Frontend/Auth/LoginRequest.php
+++ b/app/Http/Requests/Frontend/Auth/LoginRequest.php
@@ -30,6 +30,16 @@ class LoginRequest extends Request
         return [
             'email'    => 'required|email',
             'password' => 'required',
+            'g-recaptcha-response' => 'required_if:captcha_status,true|captcha',
+        ];
+    }
+
+    /**
+     * @return array
+     */
+    public function messages() {
+        return [
+            'g-recaptcha-response.required_if' => trans('validation.required', ['attribute' => 'captcha']),
         ];
     }
 }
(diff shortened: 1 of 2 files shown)


rappasoft:
Don't often see it on login screens though?

poseso:
well, at least in my country almost all systems has it, but if you think it shoudn't be in there it's ok

rappasoft:
Do you use it to help against brute force attacks or something?

poseso:
yes, specially for brutefoce attacks, i use it to prevent automated attacks

rappasoft:
Maybe it should be something that only pops up after X login failures?

poseso:
it could pop up after 3 failed attemps, then we can store the email/username and the IP address for the login attemps in a db table or count how many times the form is submitted for an IP address? and then show the captcha ?

rappasoft:
I've come up with a solution that works nicely, let me know what you think: https://github.com/rappasoft/laravel-5-boilerplate/commit/108a49817b29daa828cb479c6e2f8056cc4673a3

poseso:
very nice solution, better than what i was thinking of..

rappasoft:
It's working nice on my end. Let me know if you have time to try. I tried like 10 other ways before I got to this point.

poseso:
let me try it out and let you know soon...

poseso:
Yes, is working perfectly.

rappasoft:
Awesome. I'll merge it in.

Thanks!

PR #421, opened 2016-07-15
Small tweak for Javascript Injected from a Controller

poseso:
Translation for Javascript injected from a controller under FrontendController and strings for all languages, some languages needs translation so it will appear in english.


rappasoft:
The only reason i'm hesitant to merge this is 10/10 people are going to have to remove it from 10 files now you know? No one is going to keep that little bit of example lol.

poseso:
hahah

rappasoft:
Some things I need to let go but thanks!

vernard:
5/7 for the effort. lol

PR #444, opened 2016-07-31
Configure adminlte layout via config file

vernard:
I had to change it on my own. Probably useful if it is in config.

I copied the layout first line descriptions from the main Admin LTE theme documentation, then added my interpretation of it on the second line.

Sidebar mini was not in the documentation but it I do use it and i think it is missing from their docs.

### DIFF
--- a/config/backend.php
+++ b/config/backend.php
@@ -49,5 +49,5 @@ return [
      *
      * Note: you cannot use both layout-boxed and fixed at the same time. Anything else can be mixed together.
      */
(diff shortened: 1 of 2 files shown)


rappasoft:
Neat

vernard:
Forgot to remove the default value.

layout-top-nav looks broken to me too but it is how it looks like on the original admin lte (2.3).

vernard:
Try again. Don't use layout-top-nav.

PR #447, opened 2016-08-09
History limit

PunchRockgroin:
Add a limit to the amount of "History" taken from the database. In my application, there could be 1000 events that happen over the course of an entities lifetime, and in certain views only seeing the last 10 or so is better than 1000.

### DIFF
--- a/app/Repositories/Backend/History/EloquentHistoryRepository.php
+++ b/app/Repositories/Backend/History/EloquentHistoryRepository.php
@@ -39,10 +39,11 @@ class EloquentHistoryRepository implements HistoryContract {
 	}
 
 	/**
+	 * @param int $limit
 	 * @return string
 	 */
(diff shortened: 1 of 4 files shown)


rappasoft:
If we're adding limit do you think it would be smart to add pagination too?

PunchRockgroin:
I do. I also moved the list and item rendering functions to Blade templates.

rappasoft:
Thank you, I have taken most of this manually and added it in, and tweaked it to my liking.

Appreciate you doing the bulk of the work, this will be committed to development soon.

minedun6:
I added the plugin slimscroller and added it to the panel ^^. Should be
better than pagination i think

Le mer. 26 oct. 2016 00:55, Anthony Rappa <notifications@github.com> a
écrit :

Issue #451, opened 2016-08-15
Lazy loading on History renderType


HilmanAnshori:
Hello,

There's queries number problem on History repository when accessing renderType with string parameter.
In EloquentHistoryRepository line 63 : 

```php
$history = History::whereHas('type', function ($query) use ($type) {
   $query->where('name', ucfirst($type));
})->latest()->get();
```
and this is queries count
![screenshot_9](https://cloud.githubusercontent.com/assets/3981733/17665174/5acbd5ea-6323-11e6-808d-4161e7cd7b69.jpg)

changed it to : 
```php
$history = History::with('user')->whereHas('type', function ($query) use ($type) {
   $query->where('name', ucfirst($type));
})->latest()->get();
```

result : 
![screenshot_10](https://cloud.githubusercontent.com/assets/3981733/17665200/9813a02c-6323-11e6-8a08-9fc91d723698.jpg)

Thanks

rappasoft:
Can you make a PR?

chemitaxis:
Thanks

rappasoft:
Fixed on development thank you.

PR #452, opened 2016-08-16
Eager loading on History renderType & Ajax History 

HilmanAnshori:
Change lazy loading on History repository when accessing renderType to eager loading, as requested from [my issue](https://github.com/rappasoft/laravel-5-boilerplate/issues/451).


HilmanAnshori:
I'm also adding history limit by skip and take (not paginate) and render it using ajax call from view.

demo : 
<a target="_blank" href="http://www.youtube.com/watch?feature=player_embedded&v=JIAiJtNKkDY" target="_blank">Laravel 5 Boilerplate Ajax History</a>

BR,

rappasoft:
This is a great PR, but my only concern is what if people don't want a auto growing history section, or the history section is not the last item on the page.

What if people wanted just regular pagination with a ?history=1,2,3,4 in the url or something like that.

HilmanAnshori:
> This is a great PR, but my only concern is what if people don't want a auto growing history section, or the history section is not the last item on the page.

They can still use original function render with no parameters if they want to display all histories or 
```php
history()->render(0,10)
```
for example if they want to limit latest 10 histories. Ajax auto growing is just an example to use limiting by skip and take. And if the history section is not in the last item on the page and they want auto growing history, they can implement ajax call from js event that suite to their need.


> What if people wanted just regular pagination with a ?history=1,2,3,4 in the url or something like that.

I haven't make changes to implement regular pagination yet.  
_?history=1,2,3,4_ are history ids?

rappasoft:
history would be "page", just a different name incase theres other page parameters.

HilmanAnshori:
ok, i'll try to use paginate

rappasoft:
Basically as simple, with the lease amount of code possible. Since this is just a boilerplate.

HilmanAnshori:
```php
history()->render();
// return all records

history()->render(10);
// return latest 10 records

history()->render(false, true);
// return paginated, 10 per page, page number query string "history"

history()->render(false, true, 15);
// return paginated, 15 per page, page number query string "history"

history()->render(false, true, 15, 'pagenumber');
// return paginated, 15 per page, page number query string "pagenumber"
```
Also add renderJson to history repository for auto growing :D

rappasoft:
Thank you for this big and generous addition, however at this point I think it is too verbose to be added as a 'boilerplate' feature. I have implemented pagination by default instead.

Again thank you, maybe it will be considered in the future.

PR #478, opened 2016-09-14
added transaltion for delete role alert

motia:
The alert to confirm deleting roles did not support translation, I just appended to it the following attributes.

+                data-trans-button-cancel="' . trans('buttons.general.cancel') . '"
+                data-trans-button-confirm="' . trans('buttons.general.crud.delete') . '"
+                data-trans-title="' . trans('strings.backend.general.are_you_sure') . '"


rappasoft:

PR #484, opened 2016-09-20
Added missing DB::transaction() param

dnKaratzas:
Added missing input parameter in order to be able to send email confirmation if requested.


rappasoft:

PR #488, opened 2016-09-23
Added Dutch (nl) translation

bvbunnik:


rappasoft:

PR #496, opened 2016-10-09
Fix Html facade case in blade layouts

nayzawoo:


rappasoft:

Issue #485, opened 2016-09-20
5.3 Update


rappasoft:
Been working on a ground up since 5.3 came out. Much more changed than I thought. Probably about 30% done.

Will keep everyone posted.

digitalit:
Keep up the great work. It is much appreciated @rappasoft  :)

AlehDev:
I've been transferred to version 5.3. Later we can compare.

rappasoft:
I've started with a fresh 5.3 install. It is very much different so far. It wasn't as easy as dragging all of the folders over unfortunately.

pebbo:
Currently working on adding translatable to the roles and permissions... do you think I should wait until the 5.3 boilerplate update is ready?

rappasoft:
It's all coming together nicely, complete with new Laravel features:

<img width="1021" alt="screen shot 2016-09-23 at 11 44 41" src="https://cloud.githubusercontent.com/assets/6026504/18792094/3081f9e4-8183-11e6-9802-7aee578664fd.png">

poseso:
nice!

chemitaxis:
Thank you so much for your work...

LunarDevelopment:
Dare you give an ETA?

Amazing work so far by the way, Thankyou very much.

rappasoft:
Have a feeling this one is going to be V4 because it's changed way too much for me to not justify it.

Hoping a week or two max.

blomdahldaniel:
Thats so awesome!!

poseso:
that's great!

digitalit:
Anyone knows how this would work upgrade-wise: laravelshift.com

rappasoft:
Got stuck on the stupidest of problems, back on track just lost a few days.

rappasoft:
Since Vue 2.0 came out, I have to figure out what I need to change upgrade now. Bad timing.

blomdahldaniel:
Haha! :) That timing! Looking forward to see your work!

rappasoft:
Vue 2.0 working. BTW i'll get to the issues and PR's within this release. So when it launches i don't have to merge anything and will probably close it all.

It's getting there. Haven't had time the last couple days. But I dramatically simplified things.

I'll put it on development when i'm satisfied, but will need more eyes before I release.

minedun6:
is it possible to make something so that we can understand how to integrate another theme ? let's say metronic theme maybe or any other bootstrap theme ^^.
[...]

PR #499, opened 2016-10-12
Quick'n dirty suggestion for new dashboard and profile management

blomdahldaniel:
Suggestion from me in #485 
# This PR is not suposed to be merged
This is a dirty and quick way for me to suggest some changes around the dashboard and UX in whole. **Some restructuring has to be don but its not that much.** That would be the next thing to talk about.

Also, I think its a good idea to provide a suggested "wireframe" for the dashboard but the one I suggested here should probably be fine tuned some more first. Or maybe it should just be blank?

### Changes:
* Remove `Home` from navbar (feels redundant, instead link to `/` on logo)
* Add `Dashboard` to navbar
* Remove `Dashboard` from navbar dropdown
* Reorder navbar dropdown
* Restructure `Dashboard view` and added placehold.it as to help the developers se the potential view
  * Suggest application specific data in dashboard
  * Display small user area with link to "My profile" or "Settings"
* Add new route "My Profile" or "Settings" route:`/profile` `name('frontend.profile.index')` might be changed to `show` instead?
* create tabs for profile view. Made http-request to change the view instead of javascript
  1. Profile
  2. Change Password
* Add fa-save to buttons
skarmavbild 2016-10-12 kl 13 09 07
skarmavbild 2016-10-12 kl 13 10 27
skarmavbild 2016-10-12 kl 13 10 42


rappasoft:
Going to start adding some of this today.

blomdahldaniel:
Yummy! Let me know If you want me to do some of the parts!

rappasoft:
I actually added quite a bit of it.

blomdahldaniel:
Looks really good imo!
Just want to add "Edit Information" as a tab as well. Just like "Change Password" ?

rappasoft:
I'm not a huge fan of too many open forms on one page, I think change password is an exception.

blomdahldaniel:
That was actually why I aded the "fake tabs". So that the tabs had a separate page and then only one form on each page.

Don't you think that the whole idea falls apart if you don't go all in for consistently?
At least in my head that is not desired UX.

rappasoft:
I don't think I fully understand what you mean

rappasoft:
I took a stab it with my latest commit: https://github.com/rappasoft/laravel-5-boilerplate/commit/b130c02d4966f4b47e33d64af64d07c62cdeb8e7

rappasoft:
Let me know if i've accomplished enough to take this off my plate.

blomdahldaniel:
By just looking at the code from mobile I think it's nice and the right choice. 

Sorry for being unclear: 
In this verry PR, the tabs was actually not changed with javascript. It was changed with server reload. Good old a tag, there is bootstrap support for that functionality. So,  that way, each tab was a separate blade file = only one form per page since that was an issue for me as well! 

I dont want to put these things over you. Once v4 is released it will be easier to make PRs for things like this. I think the UX is getting to a sweet spot in this project! Glad you took your time. This is getting big man! Youve got a GitHub repo with almost 1.7k stars!! So cool actually! You can be proud man!! 😀😀 hope you can feel that in the swamp of issues.

PR #501, opened 2016-10-12
update access.php for PSR pourposes

abkrim:
Add use for Role and permission, and remove unnecesary pathson code.
If user change namespace, don't need modify this. It's PSR compilant. IMHO.


rappasoft:

PR #502, opened 2016-10-13
Fix confirmation email bug when "Send Confirmation E-mail" checkbox used on manual creation of user

ryanall:
Emails were not sent if the you had "Send Confirmation E-mail" selected on user creation as `$input` wasn't in scope.


rappasoft:

PR #503, opened 2016-10-16
Patch: MAIL_FROM & MAIL_FROM_NAME env variable added

nasirkhan:
# Enhancement
Email From and Email From Name values can be configured form the .env.


rappasoft:

PR #505, opened 2016-10-18
Update lang.blade.php Hide language

poseso:
# Enhancement for language dropdown menu

Hide current used language from the language dropdown menu, if using a language it doesnt make any sense to still see it if its currently in use.


rappasoft:
Manually added this to upcoming v4 release, thank you!

PR #506, opened 2016-10-19
fix for broken link for log-viewer when filter for All

blomdahldaniel:
So when inside `/admin/log-viewer/2016-10-19` you can click in the menu to  select filters. But when selecting **All** there is an error:
````
InvalidArgumentException in UrlGenerator.php line 314:
Route [log-viewer::logs.show] not defined.
````

This code fixes the issue but I guess it's not the right way to go.
I checked to see if this was an error with LogViewer but that is not the case.

I tried a fresh install of laravel and added the LogViewer and that worked just fine. Have been trying to find the reason to why the _All_ route does not work in this project but I could not find anything... At least this fixes it, its not perfect and not the best solution but it gets fixed.

Have a look at it if you think you'll be able to spot the mistake somewhere.

### DIFF
--- a/routes/Backend/LogViewer.php
+++ b/routes/Backend/LogViewer.php
@@ -33,6 +33,9 @@ Route::group([
 			'as'   => 'log-viewer::logs.show',
 			'uses' => '\Arcanedev\LogViewer\Http\Controllers\LogViewerController@show',
 		]);
+		Route::get('/all', function ($date){
+			return redirect()->route('admin.log-viewer::logs.show', [$date]);
+		});
 		Route::get('download', [
 			'as'   => 'log-viewer::logs.download',
 			'uses' => '\Arcanedev\LogViewer\Http\Controllers\LogViewerController@download',


rappasoft:
It's funny I couldn't figure out why this one wasn't working either.

blomdahldaniel:
Really weird! Donno what would be the right way to fix it. A total makeover is not that tempting..!

rappasoft:
Meh, i'll leave that to the package devs, we have enough to worry about.

rappasoft:
Thought this was merged, don't know why its showing as not.

blomdahldaniel:
I didnt merge it..? Wanted you to look at it first

rappasoft:
Oh, this might have been one of the ones I added manually, I thought it was one of your commits.

I'm fine with this honestly. No one will really touch the Log Viewer routes anyway theres no reason to.

blomdahldaniel:
👍

PR #508, opened 2016-10-20
font awesome icon update user -> users

nasirkhan:
Access module is for managing users, the icon will have the reflection of that :)


blomdahldaniel:
A good suggestion!

PR #511, opened 2016-10-24
Update buttons.php third time

poseso:
# Correct Translations for buttons.php Spanish

add correct translation


blomdahldaniel:
Nice! Could you take a look at the `strings.php` file as well, and add some translations to the lines under `emails` ?

Issue #515, opened 2016-10-27
User dataTables search function error 500


insic:
When try to use the search box under User management a pop apears with the following message.

DataTables warning: table id=users-table - Ajax error. 

Checking ajax response I got error 500 and the following error message.

`mb_strtolower() expects parameter 1 to be string, array given`

I use the latest boilerplate version.

blomdahldaniel:
Yes, that is true! I will try to take look at it!

nasirkhan:
It was ok in the development branch, @rappasoft might have changed something before release.

rappasoft:
Seems like a problem with the plugin: https://github.com/yajra/laravel-datatables/issues/248

But no one seems to have a fix, i'll have to trial and error it.

gotrecillo:
If we make the roles not searchable it will work.

But we wont search on the roles.

```Javascript

{data: 'roles', name: 'roles', searchable: false},

rappasoft:
Yes I just found that out too and it works that way, I was actually coming to post I don't want to do that because people will wonder why certain things aren't being searched when they can see them there.

Trying to think of a different way.

gotrecillo:
https://datatables.yajrabox.com/eloquent/joins

What do you think about this ?

rappasoft:
I'm trying to figure it out with no luck, If you can figure it out feel free to make a PR.

gotrecillo:
I got it.

Gonna clean it up and make a PR

rappasoft:
Cool thanks!

rappasoft:
We're working on this here: https://github.com/rappasoft/laravel-5-boilerplate/pull/517

If anyone can help up with the remaining bugs.

rappasoft:
Sorry this auto-closed but I want to keep it open until the bugs are fixed. I merged the search fix because it does work, but there's these issues left:

- Sorting roles column on users table duplicates rows
- Sorting permissions column on roles table duplicates rows
- Sorting or searching number of users column still errors, so I turned it off.

gotrecillo:
https://datatables.yajrabox.com/relation/morph-to-many

I think that makes imposible to sort by a column created from a relationship.

rappasoft:
Alright guess it is what it is!

PR #516, opened 2016-10-27
added tweaks to the users datatbles to allow searching

gotrecillo:

fixes https://github.com/rappasoft/laravel-5-boilerplate/issues/515

My auto code styler has formatted many things. Now that I know what to touch to make it work i could repeat it and only touch the needed lines disabling my code styler.


rappasoft:
This is a lot of change I can't exactly tell what's needed and what's not. Can you explain it?

PR #517, opened 2016-10-27
added tweaks to the users datatbles to allow searching

### LINKED ISSUE #515: User dataTables search function error 500

gotrecillo:

fixes #515

Now the changes don't affect the code style.

### DIFF
--- a/app/Http/Controllers/Backend/Access/User/UserTableController.php
+++ b/app/Http/Controllers/Backend/Access/User/UserTableController.php
@@ -47,6 +47,7 @@ class UserTableController extends Controller
 					return trans('labels.general.none');
 				}
 			})
+			->withTrashed()
 			->addColumn('actions', function($user) {
 				return $user->action_buttons;
 			})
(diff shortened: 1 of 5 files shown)


rappasoft:
Merged it in to development, but just noticed an interesting bug, try sorting the roles column. Does it add an extra row for you?

rappasoft:
Also, the roles table still errors because of the permission column, gonna try to tackle that using what you did.

rappasoft:
Ok I duplicated what you did for users on roles table.

However:

- Sorting permissions also duplicates roles
- Sorting or searching number of users column still errors, so I turned it off.
- Sorting roles column on users table duplicates rows so I turned that off too.

Maybe you can look into those since you seem to know more about this plugin than I.

PR #519, opened 2016-10-28
fix roles fetching

gotrecillo:
Now when rating a user the roles have dissapeared. This fixes that.


rappasoft:
For some reason this is causing the roles to display in the opposite order even though it is explicit about ascending the sort column.

rappasoft:
Also it breaks the search.

rappasoft:
Ok so with get, you can sort the cols but not search, without the get you can search but not sort the cols. :/

PR #524, opened 2016-10-29
FontAwesome icon added on the header area

nasirkhan:
Font Awesome Icons has been updated on the sidebar and i think the headed area has been overlooked. I use the header area this was, and am creating this PR so that other can start this way.


AlehDev:
How to be? Country Belgium 3 languages?

PR #594, opened 2017-01-02
Fix placeholder on user creation

joshbenhamou:
Incredibly minor fix. The placeholder is set to the label instead of the input on the creation form in the admin backend


poseso:
nice catch, same on password confirmation

joshbenhamou:
Good spot @poseso - had removed password confirmations so didn't notice! Updated the PR

Issue #552, opened 2016-11-12
orderBy('sort') for permissions BP Latest


digitalit:
Don't know if I'm missing something but my permissions won't sort by sort so i changed
App\Repositories\Repository on line 16 from:

return $this->query()->get();

to

return $this->query()->orderBy('sort')->get();

Is this okay or is it a better way?

blomdahldaniel:
This is interesting, if your way solves then it does that for the `sort` column only. And your way will always sort after the `sort` and I would not make such a change in the Repository!

However, this seem to have brought up a bug. The DataTables Sort By does not work for the entire Roles table. 

I tried to take a look at it but honestly I dont quite get how DataaTables are sorting things. 
When I look in the laravel debug bar I can see that the correct queries are requested. However it is not reflected in the DataTables. Almost seems like something is wrong with the javascript?
I think it is best that @rappasoft take a look at this since he knows the implementation best.

digitalit:
Well i personally don't use DataTables cause i like to have my own filtering and stuff but this were important to me since it soon gets allot of permissions and i need them to list in an correct order.

Let's say i crate an new permission for projects i want it among the other project permissions and not at the end of the list.

My fix woks for me but please tell me if you decide to solve it in another way :)

rappasoft:
Fixed in upcoming.

PR #614, opened 2017-02-02
backend role DT sorting fixed

felspire:

This is a fix for a issue #552


rappasoft:
Fixed in 5.4 release thanks!

PR #631, opened 2017-02-23
fix routes

amamarul:
Hello,
When i run composer install i had errors in routes. I fixed them
Thanks for all!
Maru


DeeJaVu:
uhm what error did you get ?

rappasoft:
Restored to old syntax for now.

Issue #609, opened 2017-01-24
Laravel 5.4


rappasoft:
I know Laravel 5.4 released today, but don't expect an update right away as I first have to wait for Laravel Collective to update their repositories to 5.4.

Also, since they changed almost everything about Vue and Webpack/Gulp/Mix whatnot, I'll probably refactor that whole thing.

Or I might just start from a fresh 5.4 installation if it becomes a problem. 

Patience is appreciated.

poseso:
👍

derekphilipau:
And I just got comfortable with 5.3 :(  Awesome job on boilerplate, take your time!

LunarDevelopment:
It's an issue in the industry when even the frameworks you adopt have
themselves shiney syndrome.

jonnywilliamson:
Thank you @rappasoft .

JPurinsh:
Hey, i was wondering, whats the progress here? :)

rappasoft:
Waiting for laravel collective to update before I start.

glaubersp:
They've created branch 7.0 for Laravel 5.4 support, @rappasoft .

rappasoft:
I've migrated most of the files over minus the tests (Dusk?) and the assets because i'm having a hard time figuring our laravel mix as it doesn't work the same as elixir and they find it necessary to change this in every version. So i'm trying to find more documentation since the laravel docs lack in this section.

derekphilipau:
@rappasoft, you could check out Spark 4 and see how they got mix working.  personally, I'm trying my best to ignore 5.4 while I learn Vue 2.

digitalit:
Also focusing on learning Vue 2 and the procedure to upgrade to 5.4 was not at all difficult BoilerPlate-wise but some of my packages i needed messed things up so im sticking to 5.3 for now.

All though i have some recommendations regarding Vue 2 and that is to start working with Axios in stead of vue-resource since 5.4 has gone over from soon to be deprecated vue-resource to Axios. 

Correct me if im wrong but here is what i did:

In bootstrap.js change 

```
window.Vue = require('vue');
require('vue-resource');

Vue.http.interceptors.push((request, next) => {
    request.headers.set('X-CSRF-TOKEN', Laravel.csrfToken);
    next();
});
```

to

```
window.axios = require('axios');

window.axios.defaults.headers.common = {
    'X-Requested-With': 'XMLHttpRequest'
};
```

and in package.json change 

```
[...]

PR #635, opened 2017-02-23
Master

amamarul:
Hello, I implemented the new Translation Strings Localization, I think you can already delete "JSON Lang Files (Maybe) - Looked at this, was much more involved than I first thought so will probably leave." In https://github.com/rappasoft/laravel-5-boilerplate/issues/609.
Also I am finishing a package to manage the strings that I hope will be uploaded soon.

It is necessary to arrange the translation of the roles in the file
"resources/views/backend/access/index.blade.php" line 35
It's working but I think we need to improve it


rappasoft:
Wow thats a lot of work thanks! I did want to do this too. I will probably take your commits and merge them in file by file instead of all at once since it's a lot though.

amamarul:
Hello!
Yes, it was a lot of work.
I was following the migration to L5.4 and I was testing, I really like your project.
When I read that you would like the Strings I saw that it was a lot and I think it is a very good new tool and very useful. For this reason I developed a package to manage the strings of the languages, I think it is getting very good, for that reason I was able to automatically generate with my package the json files.
Then I had to change all trans ('...') one by one.
I'm really glad you like my job.
I have been using BP in a project since the L5.2 version and I really like it.
I appreciate your work very much and I value it very much, for that reason I want to contribute to the measure of my possibilities.
Maybe we can do the tests in the branch development and among all we can go fixing the errors that may be. I particularly did tests but hey, you are the expert
Sorry for my English is not my native language.
regards
Maru

amamarul:
I already uploaded my package, I would love to hear your opinion. 
https://github.com/amamarul/laravel-json-locations-manager
Maru

rappasoft:
I'm going to download this and save it, it's been too long and too many changes to let github merge. We may go to json lang files in the future. Thanks for all the hard work!

PR #637, opened 2017-02-23
Rename driver name to 'pgsql' for Postgres

mikimaine:
The key for db should be 'pgsql' instead of 'postgres' to avoid the "unknown index: pgsql" Error when using Postgres as your database and execute seed command.


ragazenta:
I think db driver in TruncateTable trait should be renamed too.

**Before**
```
case 'postgres':
                return  DB::statement('TRUNCATE TABLE '.$table.' CASCADE');
```

**After**
```
case 'pgsql':
                return  DB::statement('TRUNCATE TABLE '.$table.' RESTART IDENTITY CASCADE');
```

`RESTART IDENTITY` for automatically reset sequences.

PR #638, opened 2017-02-24
Disable social logon creating accounts when registration is disabled

bryanrscott:
# FIX
<<<<<<< HEAD
When the .env attribute ENABLE_REGISTRATION is set to false new user accounts could still be created by logging in as a social user.

=======
Sorry could not work out how to split the other two small changes into a seperate request.
One is a duplicate of an existing pull request (needed to get working on postgresql)

The other is just allows a default schema to be set for postgresql in the .env

>>>>>>> release/4.0.0


rappasoft:
Thanks pulled out what was needed and added it. https://github.com/rappasoft/laravel-5-boilerplate/commit/46e531973bce579da7a8c3fcb458ae427381480c

PR #683, opened 2017-04-15
Enhancement : Add LastName / Surname to user model.

dylanharbour:
Enhancement : Add LastName / Surname to user model.

Refactor the codebase to allow for "first name" and "last name" separately. The old `name` field remains untouched but a new `surname` field is added wherever applicable. This ensures backward compatibility, although it may require users / admins to manually change current apps data slightly for best results (A scripts / migration can be written to do this automatically if needed). Naming of `name` and `surname` chosen for backward compatibility, ideally the should be called `first_name` and `last_name`. I'm open to suggestions on how this could be improved.

- Searched through all of the occurrences of 'name' with the regex `name(?!space)` and added in Surname where applicable. This takes care of models, repos, views etc.
- Updated Tests
- Updated User Factory
- Added new mutator 'getFullNameAttribute` which returns the full_name
- Updated 'access()->user()->name` usages to use the full_name
- Added the above mutator to the User Model `appends` array, This isn't required but might help some people debug or explicitly see the differences between the fields.

-TODO: Update Language Packs with "surname" where applicable.


jonnywilliamson:
For me this is great! Nice job :)

pebbo:
Updated `access()->user()->name` usages to use the full_name - hmmm, why not add firstname and surname additionally and use  `access()->user()->fullname` where you want the full name. asking, because in many project, also a "nickname" is the way to go.

dylanharbour:
@pebbo. I did this for consistency with the current code base. 
`access()->user()->name` in the current master is the same as `access()->user()->full_name` in this PR.

silverbux:
for me, i hope to just leave this to developer discretion, same as using **email** as username, currently at the project im working on using the boilerplate i simply just added a new field **username** _(for project specs reasons)_ , i think same principle applies.

jonnywilliamson:
I don't think that's a fair comparison @silverbux - here's why. If there's only a "name" field, it's like trying to unscramble an egg if you need a surname in your application...it sounds easy until you try :)

If you have a only have a name field that's been filled out by a user, then what part of the name is the firstname and which part is the surname?

So then you have to duplicate the database fields, we now have to have a `name` field, then a `firstname` field and a `lastname` field. That becomes fraught with difficulty and complications when trying to create update forms and registration forms etc.

If however the `firstname` and a `lastname` fields are only used, it becomes a win-win for both of us. It's trivial for you to continue to use a "full_name" attribute to display the name the way it currently works, yet provides a lot of flexibility for others who might need more granular control of how they address a user.

For example, if I wanted to send a message to one of my users using the current database, and assuming they are a good friend of mine, it feels wrong/too formal to start a message that will look like this:

```
Hi Timothy Johnston,

[...]
```


When I really just want to say `Hi Timothy`. By only having the full name in the database I have no reliable way of doing this.


Just my 2 cent.


Of course I appreciate that not everyone has a [firstname and lastname](http://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/) commonly found in the west.

rappasoft:
If you want, since this is a boilerplate application and i'm not worried as much for backward compatibility, you can revise to use first_name and last_name, modify the migration, etc.

I used name because it was the Laravel default, but if it's the general consensus that having them separate is necessary then i'm for it.

rappasoft:
There is also a new release that conflicts with the max:255 in this repo as they have been changed to max:191

cheshirec7:
I have made the changes as described in this PR, using name/surname. I had to make some additions as well to make everything work. I think separating the name into two parts is the right way to go.

dylanharbour:
Cool. I'll find a gap to update it as you suggested. Sounds good to me.

dylanharbour:
I've updated my PR with a new Commit 

Basically the same change as discussed above, but as @rappasoft requested I've amended all of the 'name' usages to explicitly be 'first_name'. so the User now has `first_name`, `last_name` attributes and `full_name`\`name` mutators (getters).

In cases where we were printing the `->name` method with the `access()` helper, we will now print the `->full_name` attribute. 

All tests are green (Thanks for the great test coverage!)

```
Enhancement : Add "Last Name" user model.
Refactor the codebase to allow for "first name" and "last name" separately.

Searched through all of the occurrences of 'name' with the regex name(?!space) and added in Surname where applicable. This takes care of models, repos, views etc.
Updated Tests
Updated User Factory
Added new mutator 'getFullNameAttribute` which returns the full_name. Similar mutator added for old attribute `$user->name` which will also return the full name.
Updated 'access()->user()->name` usages to use the full_name
Added the above mutators to the User Model appends array, This isn't required but might help some people debug or explicitly see the differences between the fields.

-TODO: Update Language Packs with "surname" where applicable.

**Comment by rappasoft** [comment]
Fantastic PR. Thanks for finding all of the references.

File: app/Http/Controllers/Backend/Access/User/UserController.php [hunk 1]
```diff
@@ -63,7 +63,19 @@ public function create(ManageUserRequest $request)
...
+        $this->users->create(
+            [
+                'data' => $request->only(
+                    'first_name',
+                    'last_name',
+                    'email',
+                    'password',
+                    'status',
+                    'confirmed',
+                    'confirmation_email'
```
poseso: is this confirmation_email or confirmation_code?

rappasoft: That's correct. Confirmation code should never be in the request as it's generated when it's needed. That one is this: 

screen shot 2017-05-08 at 21 50 02

poseso: 👍

PR #687, opened 2017-04-17
Update UpdateProfileRequest.php

arrakeen:
Rules for user's name must be defined in user's profile update request class as it is in registration request class

Brief enhancement of the user's validation


rappasoft:
A new PR was merged that went through all these cases.

PR #688, opened 2017-04-18
Replace glob() function usage

mikimaine:
Since glob() function have unexpected output in some systems it is probably a good idea to replace it with more consistent one (recursiveDirectoryIterator & recursiveIteratorIterator).

This is a fix. Form the official php glob() description it is known that the function have unexpected behaviors like :  

> On some systems it is impossible to distinguish between empty match and an error.

> This function isn't available on some systems (e.g. old Sun OS).

>  The GLOB_BRACE flag is not available on some non GNU systems, like Solaris.

So i have created this function with the more appropriate method to avoid this and other similar issues.

[PHP glob() doc](http://php.net/manual/en/function.glob.php)

### DIFF
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -63,7 +63,7 @@ if (! function_exists('includeRouteFiles')) {
             while($it->valid())
             {
                 if(!$it->isDot() && $it->isFile() && $it->isReadable() && $it->current()->getExtension() === 'php' ){
-                        require_once $it->key();
+                        require $it->key();
                 }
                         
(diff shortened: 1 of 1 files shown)


rappasoft:
Cool thanks!

rappasoft:
This makes the tests blow up. Can you update the tests?

mikimaine:
OK I will .

mikimaine:
@rappasoft now the test is passing. can be merge.

PR #690, opened 2017-04-18
Fix: update of Admin user causes htmlentities error

cheshirec7:
This problem occurs when a Form field is declared as an array:

{{ Form::hidden('assignees_roles[]', 1) }}

To reproduce, edit user id 1 (Administrator), then blank the name or email field, hit Update. You will get a "htmlentities() expects parameter 1 to be string, array given" error.

Form array fields must have an index to avoid this error.

{{ Form::hidden('assignees_roles[0]', 1) }}

### DIFF
--- a/resources/views/backend/access/edit.blade.php
+++ b/resources/views/backend/access/edit.blade.php
@@ -112,7 +112,7 @@
         @if ($user->id == 1)
             {{ Form::hidden('status', 1) }}
             {{ Form::hidden('confirmed', 1) }}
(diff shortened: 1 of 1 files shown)


rappasoft:
Hmm this does not happen for me?

poseso:
it happened to me and that actually fixed my problem

rappasoft:
Never mind I reproduced.

rappasoft:
Fixed thanks.

PR #692, opened 2017-04-18
Enhancement: password reset improvements

cheshirec7:
This enhancement improves the user experience when there is a problem resetting a user's password, in particular when the password reset token expires.

Currently when a user submits the password reset form and the token has expired, they get the message "This password reset token is invalid", but they are still on the password reset form, with no idea what to do next. They should be given a better message and taken to the form where they can re-send the password reset email.

### DIFF
--- a/app/Http/Controllers/Frontend/Auth/ResetPasswordController.php
+++ b/app/Http/Controllers/Frontend/Auth/ResetPasswordController.php
@@ -50,8 +50,28 @@ class ResetPasswordController extends Controller
      */
     public function showResetForm($token = null)
     {
+        $route = 'frontend.auth.password.reset';
+        if (!$token)
+            return redirect()->route($route);
+
+        $email = null;
+        try {
+            $email = $this->user->getEmailForPasswordToken($token);
+        } catch (\Exception $e) {
+        }
+
+        if (!$email)
+            return redirect()->route($route)->withFlashWarning('There was a problem resetting your password. Please resend the password reset email.'); // (invalid token)
+
+        $user = $this->user->findByEmail($email);
+
+        $exists = app()->make('auth.password.broker')->tokenExists($user, $token);
+        if (!$exists)
+            return redirect()->route($route)->withFlashWarning('There was a problem resetting your password. Please resend the password reset email.'); // (most likely token expired)
+
         return view('frontend.auth.passwords.reset')
             ->withToken($token)
             ->withEmail($this->user->getEmailForPasswordToken($token));
+
     }
 }


rappasoft:
I'm still finding it hard to merge this because what's there is the Laravel default and seems pretty self explanatory to me. Other people can weight in.

cheshirec7:
To reproduce: 

In config/auth.php set passwords.users.expire 1 (1 minute).

Now send yourself two password reset emails. Click on the first one received and you will get "An unknown error occurred", but fortunately it shows the "Reset Password" form so you can send another password reset email.

Now click on the second password reset link. Assuming the token has expired, after submitting the form it will tell you "This password reset token is invalid" and remains on the password reset form.

This PR tells the user right away before even submitting the form there is a problem with the token, so they can resend the reset email.

app/Http/Controllers/Frontend/Auth/ResetPasswordController.php
```diff
@@ -50,8 +50,28 @@ public function redirectPath()
      */
     public function showResetForm($token = null)
     {
+        $route = 'frontend.auth.password.reset';
```
dylanharbour: Shouldn't this be `frontend.auth.password.email` ?

dylanharbour:
May I recommend a refactor of this. There's too much happening in that controller method. : 

```
public function showResetForm($token = null)
    {
        //catch if the token has expired.
        if (!$this->user->nonExpiredTokenExists() ) {
            return redirect()
                ->route('frontend.auth.password.email')
                ->withFlashWarning('
                The Token you have provided is invalid or expired. 
                Please Request a new token'
                ); //@TODO: This should be bin a language Trans
        }

        return view('frontend.auth.passwords.reset')
            ->withToken($token)
            ->withEmail($this->user->getEmailForPasswordToken($token));
    }
```

with something like this in the user repo

```

    /**
     * @param $token
     * @return bool
     */
    public function nonExpiredTokenExists($token) {
        return (bool) DB::table(config('auth.passwords.users.table'))
            ->where('token', $token)
            ->where('created_at','>',Carbon::now()->subMinutes(config('auth.passwords.users.expire')))
            ->first();
    }

```

I'm not against the idea of giving  better message to the user, although I haven't found it to be an issue on. 

As long as it's not an issue, I don't mind

cheshirec7:
How about this:

```
  public function showResetForm($token = null)
    {
        if (!$token)
            return redirect()->route('frontend.auth.password.email');

        $user = $this->user->findByPasswordResetToken($token);

        if ($user) {
            if (app()->make('auth.password.broker')->tokenExists($user, $token)) {
                return view('frontend.auth.passwords.reset')
                    ->withToken($token)
                    ->withEmail($user->email);
            }
        }

        return redirect()->route('frontend.auth.password.email')
            ->withFlashWarning('There was a problem resetting your password. Please resend the password reset email.');
    }

public function findByPasswordResetToken($token)
    {
        $rows = DB::table(config('auth.passwords.users.table'))->get();

        foreach ($rows as $row) {
            if (password_verify($token, $row->token)) {
                return $this->findByEmail($row->email);
            }
        }
    }
```

cheshirec7:
resubmitted as https://github.com/rappasoft/laravel-5-boilerplate/pull/713

PR #693, opened 2017-04-19
Permanently deleting user needs updates to history log

cheshirec7:
To reproduce: Create a new user, delete, then delete permanently. In the Dashboard Recent History it will say "Admin Istrator permanently deleted user {user}" (instead of the actual user name).

This patch will show the correct user name (as plain text, not a link) for all history occurrences.


rappasoft:
This is a nice solution however it should also do the same to all past histories of that entity, otherwise clicking the link will produce a 404. So i'll get to work on that.

rappasoft:
Great PR

PR #694, opened 2017-04-20
Fix PHPUnit

poseso:
Fixes PHPUnit Which does not allow it to run.

bootstrap="bootstrap/autoload.php" 
should be: 
bootstrap="vendor/autoload.php"


rappasoft:
This is incorrect: https://github.com/laravel/laravel/blob/master/phpunit.xml#L4

I have no problem running it from that folder and that is the Laravel default.

poseso:
I do have problems running it when its not in vendor/autoload.php there is no bootstrap folder but vendor with autoload

PR #695, opened 2017-04-20
Update UserAccess.php

m-o-s-t-a-f-a:
Add some logs to find out if the given permission or role doesn't exist in database. 
It helps for debugging project specially when the spelling is not correct.


rappasoft:
Isn't this what testing is for?

rappasoft:
These methods will return false if they don't exist, if I added these messages here then people would start adding them everywhere. If they are helpful to you I think they will be something specific to your project. Thanks though.

PR #701, opened 2017-04-24
Enforce max field lengths on both front and back end

cheshirec7:
Due to a change in Laravel 5.4, as described here:
https://laravel-news.com/laravel-5-4-key-too-long-error

string fields like name and email are now defined in MySQL as varchar(191) utf8mb4_unicode_ci. 

Passing more than 191 characters to the db for fields defined this way throws a "String data, right truncated: 1406 Data too long for column" error.

This patch limits both the front and back ends to the max 191 characters for name and email fields.

Although in practice it may be unlikely to see this error as most users will not choose a username or have an email address longer than 191 characters, it should serve as a reminder that all string fields should be limited in length on the front and back end.

### DIFF
--- a/app/Http/Requests/Backend/Access/Role/StoreRoleRequest.php
+++ b/app/Http/Requests/Backend/Access/Role/StoreRoleRequest.php
@@ -27,7 +27,7 @@ class StoreRoleRequest extends Request
     public function rules()
     {
         return [
-            'name' => 'required',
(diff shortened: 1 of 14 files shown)


rappasoft:
Why did you close?

cheshirec7:
I realized I have several other similar mods to make. Almost ready to resubmit.

PR #706, opened 2017-04-27
Fix to prevent the browser from doing his own validation of the form

poseso:
Fix to prevent the browser from doing his own validation of the form using the required attribute.

In FireFox it creates undesired popup asking "Please Select an Item from the List".

### DIFF
--- a/resources/views/backend/access/change-password.blade.php
+++ b/resources/views/backend/access/change-password.blade.php
@@ -10,7 +10,7 @@
 @endsection
 
 @section('content')
-    {{ Form::open(['route' => ['admin.access.user.change-password.post', $user], 'class' => 'form-horizontal', 'role' => 'form', 'method' => 'patch']) }}
+    {{ Form::open(['route' => ['admin.access.user.change-password.post', $user], 'class' => 'form-horizontal', 'role' => 'form', 'method' => 'patch', 'novalidate' => 'novalidate']) }}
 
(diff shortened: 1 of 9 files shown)


rappasoft:
Why wouldn't you want your forms to be validated by the browser first as an extra line of security?

poseso:
Then what would be the point of using form requests? I validate via form request not via browser, I was testing out some forms of an App im making and got that undesired popup, but of course you dont have to merge this if you think its unnecessary, but IMHO i don't like that extra line of security, the popup wasn't even appearing in a real field, it was in the navigation bar

rappasoft:
What do you mean? Do you have a screenshot?

poseso:
This is correct:
correcto

poseso:
This is not correct:
incorrecto1
incorrecto2

cheshirec7:
The only way I can replicate this is to define a <select> like so:

`
{{ Form::select('test', ['' => '', 'custom' => 'Custom'], '', ['class' => 'form-control', 'required' => 'required']) }}
`
In Firefox I am seeing the "Please select an item in the list" message appear in the correct place when a <select> is defined this way.
 
Currently this boilerplate project has only two instances of <select> and neither one is using require, since there is no way to not select a value.

Seems to me using <form novalidate> would be a special case and not currently necessary in this boilerplate project.

PR #709, opened 2017-04-28
Enhancement: add redirectToHome() which depends on user login status

cheshirec7:
It seems to me that once a user is logged in their "home" page is their user or admin dashboard. 

This PR will direct users to the correct place as determined by their login status after actions occur such as role or permission failure, loginAs()  a different user, etc.

For example, when logged in as a user without backend access, attempting to navigate to /admin/dashboard should take you back to /dashboard with the "You do not have access to do that" message.

### DIFF
--- a/app/Http/Controllers/Backend/Access/User/UserAccessController.php
+++ b/app/Http/Controllers/Backend/Access/User/UserAccessController.php
@@ -37,7 +37,7 @@ class UserAccessController extends Controller
             access()->loginUsingId($user->id);
 
             // Redirect.
-            return redirect()->route('frontend.index');
+            return redirectToHome();
         }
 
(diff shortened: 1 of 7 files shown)


rappasoft:
Does this not already exist? https://github.com/rappasoft/laravel-5-boilerplate/blob/master/app/Http/Controllers/Frontend/Auth/LoginController.php#L26

cheshirec7:
I replaced all the calls that were directing users to frontend.index to go to the correct dashboard instead, unless they are not logged in, in which case they still will go to frontend.index.

I'm not sure why once a user is logged in why they would want to go back to initial landing page, but I guess they should be able to do so if they want to.

cheshirec7:
re-submitting

PR #712, opened 2017-04-29
Standardize determination of user home page

cheshirec7:
If the user is logged in their home page should be their dashboard.

### DIFF
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -98,11 +98,10 @@ if (! function_exists('getRtlCss')) {
 
 if (! function_exists('redirectToHome')) {
     /**
-     * Redirect to the "home" page depending on authentication/authorization status
+     * Redirect to the "home" page depending on authentication/authorization status.
      *
      * @param $flashMsg
      * @param $$arrData
-     *
      */
     function redirectToHome($flashMsg = null, $arrData = null)
     {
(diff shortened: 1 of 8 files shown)


app/helpers.php
```diff
@@ -93,3 +95,40 @@ function getRtlCss($path)
...
+    function redirectToHome($flashMsg = null, $arrData = null)
+    {
+        if (access()->allow('view-backend')) {
+            return redirect()->route('admin.dashboard')->withFlashDanger($flashMsg, $arrData);
+        } elseif (Auth::check()) {
+            return redirect()->route('frontend.user.dashboard')->withFlashDanger($flashMsg, $arrData);
+        }
+
+        return redirect()->route('frontend.index')->withFlashDanger($flashMsg, $arrData);
+    }
```
dylanharbour: I'm still not 100% sure I'd need this feature myself, but I'm not against the idea. 
From a code review side though, you're duplicating the `homeRoute()` Logic. I'm sure you could change redirectToHome to this and it would do the same thing. 

```
function redirectToHome($flashMsg = null, $arrData = null)
    {
        return redirect()->homeRoute()->withFlashDanger($flashMsg, $arrData);
    }
  ```

dylanharbour: I'm not sure what @rappasoft would prefer, but since this is one change request, it would be nice if it was squashed into a single commit.

cheshirec7:
Simplified and resubmitted as https://github.com/rappasoft/laravel-5-boilerplate/pull/715

PR #713, opened 2017-04-30
Password reset improvement

cheshirec7:
Currently when a user submits the password reset form and the token has expired, they get the message "This password reset token is invalid", but they are still on the password reset form, with no idea what to do next. They should be given a better message and taken to the form where they can re-send the password reset email.

To reproduce:

In config/auth.php set passwords.users.expire to 1 (1 minute).

Now send yourself two password reset emails. Click on the first one received and you will get "An unknown error occurred", but fortunately it shows the "Reset Password" form so you can send another password reset email.

Now click on the second password reset link. Assuming the token has expired, **after** submitting the form it will tell you "This password reset token is invalid" and remains on the password reset form. It should redirect to the password reset send email link form.

This PR tells the user right away before even submitting the form there is a problem with the token, so they can resend the reset email.

### DIFF
--- a/app/Http/Controllers/Frontend/Auth/ResetPasswordController.php
+++ b/app/Http/Controllers/Frontend/Auth/ResetPasswordController.php
@@ -50,8 +50,20 @@ class ResetPasswordController extends Controller
      */
     public function showResetForm($token = null)
     {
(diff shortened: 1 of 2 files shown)


rappasoft:
Merge difficulty but added thanks.

PR #715, opened 2017-05-02
Standardize home page for authenticated users

cheshirec7:
Authenticated users should return to /admin/dashboard if they have back-end permissions, /dashboard otherwise.


rappasoft:
Git is giving me problems merging this. Says patch is empty.

rappasoft:
Never mind got it.

rappasoft:
Merge difficulty but added thanks.

PR #729, opened 2017-05-12
Add contact form

cheshirec7:
Just a first stab at a Contact form, which is similar to the Registration form.


rappasoft:
This is going to be a litter verbose for what it's doing, so I will use what I can but will probably refactor a lot, but you'll still get credit.

cheshirec7:
Going to resubmit with just the files actually needed.

PR #730, opened 2017-05-12
Add contact form

cheshirec7:
Very similar to the Registration form.


rappasoft:
Added one myself but thanks for the request!

PR #733, opened 2017-05-15
Fix first and last name bug on social logins. 

daniel-farina:
Social providers do not provide first and last name all the time. Google will put it here:
$data->user['name']['familyName'];
$data->user['name']['familyName']

while Facebook only provides: 
 $first_name = $data->user['name'];

This could be why Laravel only keeps only a name column instead of first and last name. 
 This code temporally fixes this issue.


dylanharbour:
I don't use Facebook / Google Auth, so I actually never thought of this when I refactored. Thanks for picking it up @daniel-farina !

daniel-farina:
Absolutely, 
Thanks for adding the first & last name. It's something I had in my mind too. 

Cheers!

rappasoft:
Did you test these?

rappasoft:
Used this one because it was tested. https://github.com/rappasoft/laravel-5-boilerplate/pull/734

But thanks as always.

PR #744, opened 2017-05-26
Add ability for admins to manually approve accounts

clin407:
Added new configurations flags that disable the ability to automatically send emails
for users to confirm their account. Useful for when the admin wants to manually
approve each account. 

When confirm email flag is set to false, and the new flag is set to true, it will still show a pending message to users.


rappasoft:
I ended up taking a lot of what you had, but it needed a lot more so I added it:

demo

clin407:
sounds good to me!

PR #745, opened 2017-05-26
Changed default permission in Request, override failedAuthorization

spgwebdev:
The old method `forbiddenResponse` was replaced with `failedAuthorization` as per [commit](https://github.com/laravel/framework/commit/1a7540967ca36f875a262a22b76c2a094b9ba3b4).

### DIFF
--- a/app/Http/Requests/Request.php
+++ b/app/Http/Requests/Request.php
@@ -2,8 +2,8 @@
 
 namespace App\Http\Requests;
 
-use Illuminate\Foundation\Http\FormRequest;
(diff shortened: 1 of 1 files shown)


rappasoft:
What is this doing again?

spgwebdev:
It replaces the old method, but it overwrites the new method from Laravel (please see [commit](https://github.com/laravel/framework/commit/1a7540967ca36f875a262a22b76c2a094b9ba3b4) ) in case of an unauthorized request throws an error.

PR #777, opened 2017-06-28
misc updates

cheshirec7:
Some suggested changes I am using in my own projects. Mainly blade templates should be using {!! !!} instead of {{ }}, especially for trans() calls, since {{ }} calls htmlspecialchars() which will, for example, replace single quotes with & #39; in the translation strings.

The most visible changes are for the user and role datatables. 
Note the custom spinner that appears during all ajax calls. 
Also, flash messages can be dismissed.
 
A project with all these changes is available here:
http://boiler.erictotten.net/


rappasoft:
{{ }} are important for security, the only thing I use {!! !!} on right now are any attributes that output html. At one point we had everything {!! !!} but we changed it to use {{ }} just in case of sql or js injection. Will have to look into it.

rappasoft:
I like this PR but I don't agree with the {{ }} to {!! !!} change. The !! in the second one is suppose to signify that it is dangerous to use.

cheshirec7:
I agree, fields that come out of the database like {{ user->first_name }} need XSS protection, and I still have those using {{ }}. However, almost everything else like {!! trans('labels.backend.access.users.table.first_name') !!} are safe - unless someone can hack the language files? Also note that calls like {!! Form::label('first_name', trans('validation.attributes.backend.access.users.first_name'), ['class' => 'col-lg-2 control-label']) !!} already get run through htmlentities() by default.

rappasoft:
Cherry picked a few things in the latest release: https://github.com/rappasoft/laravel-5-boilerplate/commit/bc18b626401f9ffcf894af5a18ee536cc126fdc8

Not crazy about the blade stuff but if you wanted to do a PR with just the DT portion I would take it.

cheshirec7:
will be resubmitting with changes broken out

PR #795, opened 2017-07-27
Provide basic skeleton for API routes

bryanrscott:
# enhancement?
While not providing any api routes it provides the template and folder for adding api routes.    Makes it easier if someone wants to add api routes?

### DIFF
--- a/routes/api.php
+++ b/routes/api.php
@@ -16,5 +16,5 @@
  * Namespaces indicate folder structure
  */
 Route::group(['namespace' => 'Api', 'as' => 'api.'], function () {
-    includeRouteFiles(__DIR__ . '/Api/', false);
+    includeRouteFiles(__DIR__.'/Api/', false);
 });


routes/Api/README.txt
```diff
@@ -0,0 +1 @@
+API routes can be placed here.
```
bloggervista: I guess instead of readme it is better to use 
User.php to display user details with auth middleware.

routes/api.php
```diff
@@ -13,6 +11,10 @@
 |
 */
 
-// Route::middleware('auth:api')->get('/user', function (Request $request) {
```
bloggervista: How about adding it to inside API folder you created?

routes/Api/README.txt
bryanrscott: I was not sure how far @rappasoft wanted to take this.   Was just doing the bare minimum while still making it easier.

routes/api.php
bryanrscott: I am thinking the intent is to create a minimal prep for supporting API's in a similar folder structure to the rest of boilerplate without impacting to much.   The line was commented in boilerplate.

rappasoft:
The reason I didn't include any API stuff is everyone has their own way of doing it. I'll look this over and maybe I can come up with a compromise.

routes/api.php
bryanrscott: Do you mean as a comment like the original.  Yeah I think that would be an improvement

bloggervista: @bryanrscott  yea :)

bloggervista:
@rappasoft  I guess with laravel 5.5 the most common way to do api is by using new Resource isn't it?

rappasoft:
Closing this for upcoming 5.5 release where it's different.

PR #796, opened 2017-07-27
translate root breadcumbs

tamplan:
Hi,

Fix root breadcumbs element to be translated.


tamplan:
Update french language files

resources/lang/fr/buttons.php
```diff
@@ -18,14 +18,14 @@
             'users' => [
                 'activate'           => 'Activer',
                 'change_password'    => 'Changer de mot de passe',
-                'clear_session'         => 'Clear Session',
-                'confirm'             => 'Confirm',
+                'clear_session'       => 'Effacer la session',
```
bloggervista: Alignment with other => I mean this line has different alignment than line 22, 21

tamplan: Hi,
I do not feel still comfortable in use of git for the moment.
Did i need to do some changes?

Regards.

bloggervista: Oh then no , not needed now. I will later change it :)

tamplan: Ok, thanks you very much!

rappasoft: Don't worry about formatting, StyleCI has an auto hook that formats when I pull to dev.

PR #810, opened 2017-08-24
Added Function renderByUser

vangelis183:
Enhancement:
Added function `renderByUser`
Passing the User-ID as last parameter gives you the Option to render History for all Entities by User


rappasoft:
History is being removed in upcoming release and replaced with spatie/activitylog, but thanks for this.

PR #817, opened 2017-09-01
add updated zh translation

kiwikefei:
update the zh translation to the latest version.


rappasoft:
Going to merge these files manually into 5.5 and close this when done.

PR #827, opened 2017-09-13
Sidebar menu

ccsliinc:
This PR is for an enhancement for the side-bar menu.  Instead of using html and blade this allows the menu to be created with with the spatie/laravel-menu plugin.  All the menu code is neatly packaged into a Menu Service, which allows for even more options to build menu.  Included is the original menu exactly as was, recreated with the plugin.  For information regarding the menu plugin please see.

https://docs.spatie.be/menu/v2/introduction#.


ccsliinc:
Opening a new one

PR #829, opened 2017-09-13
Added Menu Plugin

ccsliinc:
This PR is for an enhancement for the side-bar menu. Instead of using html and blade this allows the menu to be created with with the spatie/laravel-menu plugin. All the menu code is neatly packaged into a Menu Service, which allows for even more options to build menu. Included is the original menu exactly as was, recreated with the plugin. For information regarding the menu plugin please see.

https://docs.spatie.be/menu/v2/introduction#.


rappasoft:
This seems like a whole lot of files and code to render a sidebar don't you think?

ccsliinc:
I believe half the files are from compilation.  It's a menu plugin with built in active support.  Creating new menus can be done via php vs blade which is where is belongs, especially with mixing of authorization and isactive.  The full build out was to match your exact menu structure, and to make it self explanatory for others to modify their structure.  I understand where you are coming from but if we break it down it's not much more added in terms of files, but much cleaner code?  Thoughts?

ccsliinc:
I do apologize, I guess many of those files should not have been added, it seems they are not in the ignore which i guess they should be.

rappasoft:
I like the PR but I think this is something people could add it they want it. Thanks.

PR #848, opened 2017-09-24
Include pace js

nasirkhan:
installed Pace directly from the github by adding "pace": "github:HubSpot/pace#v1.0.2" in pacage.json. 

`yarn build` passed successfully. but shill not showing.

### DIFF
--- a/app/Http/Controllers/LanguageController.php
+++ b/app/Http/Controllers/LanguageController.php
@@ -2,6 +2,8 @@
 
 namespace App\Http\Controllers;
 
+use Illuminate\Support\Facades\Session;
+
 /**
  * Class LanguageController.
  */
@@ -14,7 +16,10 @@ class LanguageController extends Controller
      */
     public function swap($locale)
     {
-        app()->setLocale($locale);
+        if (array_key_exists($locale, config('locale.languages'))) {
+
+            Session::put('locale', $locale);
+        }
 
         return redirect()->back();
     }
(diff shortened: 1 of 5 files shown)


rappasoft:
Hmm but if you run npm run dev you still get build errors. Thats the problem.

nasirkhan:
i am not getting any error now. `npm install` again and `npm run dev`. if you check the commit i did not change much.

rappasoft:
I get this error, even though the actual file exists in node modules: 

screen shot 2017-09-25 at 09 29 25

bloggervista:
i guess it is due to require('../../../../node_modules/pace-js/pace.min');  cannot we do ./node_modules/pace-js/pace.min

rappasoft:
I tried both, same error. File exists I don't know why it's not picking it up.

nasirkhan:
i was getting the same error and then googled a bit, and found that added "
pace": "github:HubSpot/pace#v1.0.2", and then 'yarn' and 'yarn dev' and it
stopped showing the error. I will check again and inform you the updates.




--
*Nasir Khan Saikat*
www.nasirkhn.com

nasirkhan:
i am not getting any error. removed the `node_modules` and then ran the `yarn` and `yarn dev`. build passed successfully.

bloggervista:
nasirkhan you may be from windows. It may be cross platform issue.

Rappasoft lets not mind this error . We can download pace.js and use mix.copy for now and we can think its solution later.

nasirkhan:
@bloggervista i am not running windows. Just google the error and you will find that a lots of other people faced the same issue. 

You are right it is not a show stopper and we can add this later. 

for discussing the upgrade plan we were using the slack channel (https://laravelboilerplate.slack.com/messages/C75A9UJBC/), lets use that more actively.

rappasoft:
Fixed thanks!

PR #855, opened 2017-09-28
 Fix log viewer and added page-header

Shuyinsama:
### Please double check before merging

I have changed the blades for Log Viewer and added a simple yet effective (same as old BP) page-header section to the blades.

As mentioned in my last commit to this branch:
Because there are a lot of irregularities between BS4 and CoreUI/LogViewer styles, it so happens that a lot of Log Viewer text becomes unreadable and colours are overwritten by BS4.

This meant that I had to add a seperate `_log_viewer.scss` file to override some styles.
Sadly because Log Viewer html does not indicate it is Log Viewer it was not possible to target those pages/elements specifically.

**Question:** Would it be best to add classes to the Log Viewer Blades to target those elements specifically instead of global style changes?


rappasoft:
Yeah it would probably be best to add custom classes so they don't interfere with anything else.

PR #862, opened 2017-10-02
Permissions field needed to be checked as it is optional

nasirkhan:


rappasoft:
I botched the merge on this so i'm gonna close, but its in there.

PR #868, opened 2017-10-05
Removing unsigned usage with increments

kt-12:
Use of unsigned() with increments() is redundant. increments() itself create a auto increment unsigned integer.

refs: ( $table->increments('id');  under Available Column Types )
https://laravel.com/docs/5.5/migrations#creating-columns


rappasoft:
The new branch doesn't have 90% of these tables, but I fixed it on the one that had it. Thanks.

PR #875, opened 2017-10-08
new UI for Users and Roles pages

nasirkhan:


rappasoft:
Can you do me a favor and only commit the view files so it doesn't conflict. 

Also I don't think we need the orange back buttons since the cancel buttons do the same thing.

As well as the responsiveness seems a bit off.

Thanks!

nasirkhan:
pushed new updates. no conflict now. 
Can you please mention where you found responsiveness issue, i can update those pages also.

PR #886, opened 2017-10-16
Added Boostrap 4 templates as default pagination view

matthewnw:
Changed the default Laravel AbstractPaginator class to use the built in Bootstrap 4 blade templates.
Moved pagination out of the Bootstrap 'card-footer' as it was causing the pagination to display as table.


matthewnw:
I can't figure out what StyleCI is asking to change?

PR #890, opened 2017-10-19
Fix removing all permissions and add CoreUI Switches

matthewnw:
- Fixed a bug where all permissions could not be removed from a User or Role (even not required in the app config option). 

- Fixed label for attr for email confirmation on User creation

- Added CoreUI switch styling to checkboxes on backend Users and Roles.


rappasoft:
Merged but it didn't register so just closing. Thanks.

PR #896, opened 2017-10-22
Admin index. (redirect)

lloricode:
### Fix
Additional route for admin index,

### DIFF
--- a/routes/backend/admin.php
+++ b/routes/backend/admin.php
@@ -3,5 +3,5 @@
 /**
  * All route names are prefixed with 'admin.'.
  */
-Route::redirect('/', '/admin/dashboard', 302);
+Route::redirect('/', '/admin/dashboard', 301);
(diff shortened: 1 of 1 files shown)


rappasoft:
How about just a redirect?

lloricode:
@rappasoft it make sense if we add a route/url for index in admin to easy viisit in admin like
`myweb.com/admin`

lloricode:
just like a normal frontend index
```php

Route::get('/', 'HomeController@index')->name('index');
Route::get('dashboard', 'DashboardController@index')->name('dashboard');

```

rappasoft:
Right what i'm saying is just redirect /admin to /admin/dashboard instead of creating more controller methods and breadcrumbs

Route::redirect('/here', '/there', 301);

lloricode:
hmm, new idea to me, i try now

lloricode:
@rappasoft, is this right?

PR #944, opened 2017-11-09
Small change (comments) in index.blade.php

sbreiler:
Its a polish (enhancement?) for /resources/views/frontend/index.blade.php:
End-comments for columns said "col-md-10", but they are "col-xs-12" - changed the comment to fit.

### DIFF
--- a/resources/views/frontend/index.blade.php
+++ b/resources/views/frontend/index.blade.php
@@ -17,7 +17,7 @@
                 </div>
             </div><!-- panel -->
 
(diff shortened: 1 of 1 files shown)


Shuyinsama:
This happens in a couple of different places too. Could you look over these too if possible?
On the other hand. A new 5.5 release is close, besides it being a cleanup is it still needed?

rappasoft:
Will wait for 5.5 as this would get overridden.

PR #963, opened 2017-11-28
Updated .env.example social default values

jericopulvera:
This pr improves the default .env.example social credentials for easier socialite setup.


rappasoft:
Why are you turning them all on though?

jericopulvera:
It is commented so it is the same as false.

rappasoft:
Ah true true ill merge on next release

jericopulvera:
Do you think it's better to just remove the comment and set the active value to false. 

What do you think?

PR #964, opened 2017-11-28
made last_name column nullable

jericopulvera:
Fix for social accounts that does not have a last name.


rappasoft:
This would require more than just nulling out the field. Which providers don't give last name?

jericopulvera:
twitter

rappasoft:
git is giving me a hard time merging this for some reason so i'm gonna do it manually. Actually going to null out first name and last name and just figure out which are needed using form validation.

PR #970, opened 2017-12-03
Added doctrine/dbal ^2.6 to allow schema updates

christoph-kluge:
Right now you're not able to perform "schema" updates because you get an exception:

```
 [RuntimeException]                                                                   
  Changing columns for table "users" requires Doctrine DBAL; install "doctrine/dbal".
```

According to laravels documentation ([migrations#modifying-columns](https://laravel.com/docs/5.5/migrations#modifying-columns)) it's required to have `doctrine/dbal` installed.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -5,7 +5,7 @@
     "license": "MIT",
     "type": "project",
     "require": {
-        "php": ">=7.0.0",
+        "php": ">=7.1.0",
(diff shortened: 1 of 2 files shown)


rappasoft:
Is it necessary to have PHP 7.1 for this?

rappasoft:
I'll include the package but keeping php version at the current laravel version default.

PR #972, opened 2017-12-03
Fixes google's no captcha validation

christoph-kluge:
As soon as you try to submit the registration form with activated NoCaptcha then you'll get an missing member-Exception because NoCaptcha's validation-rule usage for Laravel 5.5 got changed.  Additionally it's not required to register the alias anymore.


christoph-kluge:
Can someone tell me maybe why the StyleCI wants me to put the namespace declaration in the middle of two existing (alphabetically correct placed) declarations? :/

PR #987, opened 2017-12-22
Countable fix

SketchNI:
Change count() to null !== func() for scopes, with and fields in SocialLoginController::getAuthorizationFirst().

It was causing a Countable error. Related notes: https://wiki.php.net/rfc/counting_non_countables

phpterm


rappasoft:
Shouldn't all these be arrays though even if empty?

rappasoft:
Closing this until someone tells me why this is an issue if you keep them empty arrays.

Issue #979, opened 2017-12-06
v5.5 Testing


phpgraham:
Clean install no changes yet, other than creating Model with:
php artisan make:model Pages -mr
created a route and view, which works fine in browser
then created a simple test, using php artisan make:test

namespace Tests\Feature;
use Tests\TestCase;
use Illuminate\Foundation\Testing\RefreshDatabase;

class PagesTest extends TestCase
{
    /**
     * A basic test example.
     *
     * @return void
     */
    public function testExample()
    {
        $this->assertTrue(true);
    }
}

Get error:
Fatal error: Class 'Tests\TestCase' not found in C:\Homestead_Projects\laracmsnew\tests\Feature\PagesTest.php on line 8
PHP Fatal error:  Class 'Tests\TestCase' not found in C:\Homestead_Projects\laracmsnew\tests\Feature\PagesTest.php on line 8

Line 8 is:  class PagesTest extends TestCase

This simple test works fine on a clean laravel 5.5 install on same server with same install procedures.

Is Tests\TestCase extends Illuminate\Foundation\Testing\TestCase  working on BoilerPlate  or is it just BrowserKitTesting that works, the 92 tests that come with BoilerPlate all work

How can i run phpunit tests ?

SketchNI:
Hi. I followed your steps and got the same error your did

`PHP Fatal error:  Class 'Tests\TestCase' not found in /home/sketch/Code/L5B/tests/Feature/PageTest.php on line 6`

I looked in `TestCase.php` and it doesn't have a namespace so I removed that declaration and still had the same issue. I looked at some of the other tests and saw they were using BrowserKitTesting and changing `class PageTest extends TestCase` to `class PageTest extends BrowserKitTestCase` allows the tests to run successfully.

My PagesController@index just returns "OK" to the browser so writing this:
````php
<?php namespace Tests\Feature;

use Illuminate\Foundation\Testing\RefreshDatabase;
use Tests\BrowserKitTestCase;

class PageTest extends BrowserKitTestCase
{
    /**
     * A basic test example.
     *
     * @return void
     */
    public function testExample()
    {
        $this->visit('/pages')
            ->see("OK")
            ->assertTrue(true);
    }
}
````
allows me to confirm that BrowserKitTestCase works.

bambamboole:
This is an issue with the test setup.
I made an PR with a refatocring of the whole test suite to Laravel 5.5 Conventions.

SketchNI:
[...]

PR #992, opened 2018-01-05
Update Test Suite and remove Browserkit Dependency

bambamboole:
Hi @rappasoft,

first of all, this is a very nice project. I really appreciate you

I'm new to open source and read your message that it is hard to maintain such a package.
So maybe I can help you a bit :-)

This PR will refactor the Test Suite to latest Laravel Conventions and remove Browserkit Dependency.
 

This PR add a line in `phpunit.xml` that the `MAIL_DRIVER` is set to `log` that the tests do not fail initially.

Also I changed the language switch route to not call the method `swap` on the `LanguageController` but invoke it via `__invoke`.

cheers
bambamboole


rappasoft:
Why don't you like browser kit? It makes it easier to understand what's going on and faster overall to make tests.

bambamboole:
I think it would be cleaner to use built in stuff. This is also the way the official laravel docs describe how to write tests. 
I would also recommend to sort tests in feature and unit tests and resort them a bit for a better overview.
The result would be an updated easy to extend test suite.

bambamboole:
So until now All Tests are refactored to use the default TestCase and Browserkit as dependency is removed.

bambamboole:
This PR also solves Issue #979 .

rappasoft:
You can go ahead and do what you stated above, after that I will give it a thorough look over. Good work.

bambamboole:
@rappasoft , I've refactored all tests.
Now there are less tests and assertions. They are more explicit and (in my opinion) better sorted, because each Feature has its own Test Class.

Another change is, that I avoided to assert against to see something on the page, because changing the layout would maybe break the tests. Instead I asserted against the flash messages in the session and used the translations so that there was no need to set a language in the tests.

Additionally I never used the setUp method, so each tests builds up the stuff its need. I only declared a loginAsAdmin method on the TestCase because that would be some kind of cumbersome to write this in each test.

There are not as much tests like before, but I also think they are faster than before.

At least I have to say that I think not every use case is tested, but almost. While refactoring the test case I got a very good insight of the code. Really looking forward to make more PR's to help you.

rappasoft:
This is crazy. Can't wait to review it all. If you have time I could use help with the issues list too. Would be appreciated.

bambamboole:
@DrPandalot thanks!

I think we don't need an .env.testing file because all envvars which need to be changed for the tests are declared in phpunit.xml.

I fixed the broken tests. I was sure that all tests were passing... Now all are passing :-)

bambamboole:
The last commits were an accident while playing with Jenkins. Sorry for that...

PR #993, opened 2018-01-09
evade stash

trollid:
this will stop :
``
Please, commit your changes or stash them before you can merge.
Aborting
``

enchancement

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,8 @@
 /node_modules
 /public/hot
 /public/storage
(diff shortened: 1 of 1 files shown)


bambamboole:
the composer.lock file needs to be in the repository because it tracks the specific versions that composer installs.

The other files are maybe for convenience commited, that you don't have to build the frontend after git clone.

rappasoft:
I think these are all important to have.

PR #1000, opened 2018-01-12
Enhancement: Add install command to streamline the installation process

bambamboole:
This PR adds an boilerplate:install command, which does multiple things:

It generates  an application key and asks for database credentials which are filled in .env file.

Tests are covering the main functionality


rappasoft:
I actually used to have an installer that did the following:

- Download
- Composer Install
- NPM Install
- Run Gulp
- Connect Database
- Migrate
- Set Admin Account
- Seed
- Set Namespace

But every time I made a big addition something would change and break it and I didn't have time to keep it up. Maybe we can add some more functionality to this one instead. I like that it's a command and not a separate repository.

bambamboole:
First there was npm stuff with in it, but because of the committed built assets I removed it.
I'm working on a pr which removes the compiled assets from the repo.

We can't composer install in the command because it won't work without installed dependencies.

I had some issues with migrating database because we have change config values on the fly.

But I will continue working on it

rappasoft:
I'm just not sure its worth the added effort if it does what I can do in a few seconds. Anyone that's using this should know how to set up a default laravel application anyway.

rappasoft:
Gonna close this for now.

PR #1001, opened 2018-01-12
Enhancement: Update Core UI to 1.0.6

### LINKED ISSUE #975: Question: CoreUI version outdated?
General question, what version of CoreUI have you integrated into your 5.5 release?

I only ask as I see references to v1.0.0-alpha.6 in the CoreUI resource files which appears to be the projects' initial release from October with several releases since:

https://github.com/mrholek/CoreUI-Free-Bootstrap-Admin-Template/releases

bambamboole:
I just compared the swapped out the sass files and updated yarn.lock.

Can't see any major changes or broken things in the backend so this should just work.

I saved the logo in resources and removed the image in public. 
We should avoid to commit any resources that can be build with laravel mix.

btw: this resolves #975


rappasoft:
There seems to be other things in here not directly related to coreui, like removing pace and global overrides? You need to explain why so I have something to link to when people ask.

bambamboole:
@rappasoft 
I tried to use pace-progress in favor of pace, because with pace-progress there would be an npm package. But there is an issue with it [(click me)](https://github.com/HubSpot/pace/issues/328) .

So I built it back but changed the require parameter because there doesn't have to be the whole path to node_modules. the package name is enough.
The reason for a new yarn.lock was the yarn update command.

PR #1004, opened 2018-01-17
Enhancement: add circle ci config

bambamboole:
We should leverage continuous integration in this project.
A big Pro of this is that we could see in Pull Requests if anything is breaking.
As an open source project it's also good to signalize any visitor that this project is backed by ci with a badge.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,34 @@
+version: 2
+
+jobs:
+  build:
+    docker:
+      - image: circleci/php:7.1-node-browsers
+    working_directory: ~/laravel
+    steps:
+      - checkout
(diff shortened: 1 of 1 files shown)


rappasoft:
I've never used it, comment (for me and everyone) how they would go about setting it up if you have time. So I can test it for one. And also a brief synopsis of what it's doing. Thanks

bambamboole:
@rappasoft 
Circle CI is an continuous integration service which runs on each push to master and on each pull request.

The only thing you have to do is to login with github on https://circleci.com .
It looks for the config file which this PR adds and then runs phpunit.

It will fail in the current master because of the current test setup.

I would prefer a merge of the Refactor Test Suite PR before merging this because then I will check that the CI is passing when merging this PR.

If you have any other questions, just ask :-)

PR #1008, opened 2018-01-23
Set default timezone form .env when creating new users

wizston:
This is an enhancement to preselect the timezone dropdown from the .env file that the user has set up initially instead of "hardcoding" the default to "UTC".


bambamboole:
@wizston You should take `config('app.timezone')` instead of using environment variable.

wizston:
@bambamboole Oh yeah, thanks for that, I've made changes accordingly.

PR #1012, opened 2018-01-26
Enhancement: Add Frontend test suite

bambamboole:
This PR adds a frontend test suite based on facebook/jest with a first test for the ExampleComponent.
It also triggers test runs in circle ci.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,35 @@
+{
+  "env": {
+    "test": {
+      "presets": [
+        ["env", {
+          "targets": {
+            "node": "current"
+          }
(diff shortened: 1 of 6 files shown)


rappasoft:
Can you just describe all that's going on in there?

bambamboole:
jest does not need webpack or laravel-mix to work. because of that we need a little bit of configuration. Thats why we need a .babelrc file and the jest section in the package.json.

Additionally I added two commands in package.json for running tests and running test with watcher.
At least there is a Test for the Example Component to show how the testing works.

PR #1013, opened 2018-01-27
Remove compiled assets

bambamboole:
This PR remove compiled assets from the repository.

They are not needed in the repository because the can be built automatically.
On changed the have big diffs and bloat up the whole repository and can lead to merge conflicts.

Additionally I removed tether.js because bootstrap need popper.js since beta and removed unused variables from app.js


rappasoft:
screen shot 2018-01-30 at 09 43 02

I'm with him on this, I understand both sides, but there need's to be a plan for people if they're not being vc'd. I know I don't want to install and run npm on my servers and have 100k extra node_module files...

bambamboole:
@rappasoft I know what you mean. But when you work with at least two persons on a project you will have diffs everytime you build the assets. But I will tackle some solutions for that and share them with you. Just keep it opened so other contributors will see we are discussing on it :-)

rappasoft:
Needs more thought

PR #1016, opened 2018-01-29
Fix for getPicture Method where storage folder is missing

poseso:
storage/ is missing from the returned url in the getPicture method


bambamboole:
@poseso Thanks a lot for the fix :-)

rappasoft:
This doesn't seem to be working for me.

PR #1031, opened 2018-02-21
Fixed minisidebar dropdown submenu nav-link padding

poseso:
This fixes the padding on the dropdown submenu when minisidebar is active


poseso:
screen shot 2018-02-21 at 6 00 29 pm

PR #1032, opened 2018-02-25
add formatter

lloricode:
Add command
`composer format` to format file and `composer test` to run phpunit


bambamboole:
oh that formatter is nice. but do we need the php-cs cache file in versioning? I think it would be better to add it to the gitignore.

PR #1038, opened 2018-03-05
Add simple search to the user admin pages.

bryanrscott:
# enhancement?
Add simple search to the user admin pages.

### DIFF
--- a/app/Http/Controllers/Backend/Auth/User/UserController.php
+++ b/app/Http/Controllers/Backend/Auth/User/UserController.php
@@ -41,7 +41,8 @@ class UserController extends Controller
     public function index(ManageUserRequest $request)
     {
         $search = $request->get('search', '');
-        $orderBy = $request->get('orderBy', 'id');
+        //$orderBy = $request->get('orderBy', 'id');
+        $orderBy = 'created_at';
         $sort = $request->get('sort', 'desc');
         $users = $this->userRepository->getActivePaginated(25, $search, $orderBy, $sort);
         $users->appends(Input::only(['search', 'sort', 'orderBy']));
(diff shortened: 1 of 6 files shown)


bambamboole:
@bryanrscott thats really nice, but can you add some tests to prove the functionality?

Shuyinsama:
Nice pull.

As a counter argument. Is it really handy to have such a functionality embedded inside the boilerplate?

I've made the backend search available in multiple project by adding Algolia search which basically only requires adding Laravel Scout and add a trait to the Model that you want to be searchable.

I think that way, even though it requires a external party, it's way more versatile in terms of boilerplate code.

With these changes something like that would be way more difficult to add since some of the code changes are in a lot of core files.

Maybe @rappasoft has a view on this too?

bambamboole:
@Shuyinsama a scout implementation would be more useful. But I think there is no useful MySQL implementation of Scout. I found [this](https://github.com/YABhq/laravel-scout-mysql-driver) package, but I think it is not mature enough because of all the issues.

rappasoft:
We removed this functionality from the last version along with datatables since I wanted people to be able to use their own implementation and not have me force plugins on them. I haven't looked at this yet but it might be the same thing we removed in the first place.

bryanrscott:
This was intended to add search in a minimal way, so that out of the box the boilerplate is more useful.   It is limited because no plugins or modules where added.

@Shuyinsama the difference is adding a search parameter to the  

@rappasoft No plugins added.

bryanrscott:
@Shuyinsama I moved the fixing of orderBy back to the controller, after thinking about your comments.  This is also where it is done in the original

@bambamboole I will add tests if decided to proceed with

rappasoft:
Appreciate the work, but going to hold off on this for now because I may want to add some sort of datatables plugin back in.

PR #1048, opened 2018-04-04
Language pack for Iranian (Persian)

htaghizadeh:
Language pack for Iranian (Persian)


htaghizadeh:
Unfortunately, I forgot to add Persian (fa) language folder in the lang folder!

PR #1056, opened 2018-04-11
Modify role update logic

wgqi1126:
Modify role update logic, allow change role name's uppercase and lowercase

### DIFF
--- a/app/Repositories/Backend/Auth/RoleRepository.php
+++ b/app/Repositories/Backend/Auth/RoleRepository.php
@@ -113,8 +113,9 @@ class RoleRepository extends BaseRepository
     {
         $query = $this->model
             ->where('name', $name);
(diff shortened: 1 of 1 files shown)


app/Repositories/Backend/Auth/RoleRepository.php
```diff
@@ -105,13 +105,17 @@ public function update(Role $role, array $data)
...
+    protected function roleExists($name, $exclude = null)
     {
-        return $this->model
-                ->where('name', $name)
-                ->count() > 0;
+        $query = $this->model
+            ->where('name', $name);
+        if ($exclude !== null) {
+            $query->where('id', '!=', $exclude->id);
+        }
```
bingalls: Putting the $query in an else ($exclude === null) clause prevents it from running twice

PR #1063, opened 2018-04-27
5.1.5 css

bingalls:
## Sass Testing
Adds
* sass-lint
* stylelint
Fixes *.scss code to pass these tests
Adds editorconfig & testing.md to explain these tests

### DIFF
--- a/testing.md
+++ b/testing.md
@@ -1,6 +1,6 @@
 ## Testing
 * `cd laravel-5-boilerplate/`
-* `yarn add --global sass-lint stylelint`
+* `npm i -g eslint jscsrc jshint sass-lint stylelint`    # or yarn global add 
 * `composer validate`
 * `composer phpunit`
 * `composer coverage-text`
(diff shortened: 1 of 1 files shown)


bambamboole:
whats the difference between sass-lint and stylelint?

bingalls:
@bambamboole sass-lint & stylelint have an intersection, but catch different things:
https://github.com/bjankord/stylelint-config-sass-guidelines/wiki/Lint-Report-Comparison
Note that this compares the deprecated scss-lint. However, I believe that it is a fork of sass-lint, and most of its configuration still maps to sass-lint.
Here is a graph on the maintenance performance of the 2 projects:
http://www.npmtrends.com/known-css-properties-vs-sass-lint-vs-stylelint
stylelint can also check css, as well as other tools. We also should soon look at Prettier. Here's an overview from CSS guru Chris Coyier:
https://css-tricks.com/prettier-stylelint-writing-clean-css-keeping-clean-code-two-tool-game/

PR #1072, opened 2018-05-16
add missing vue dependency in backend

damieneskimo:
# Enhancement
This PR is to add Vue as a dependency in the backend, which can allow users to use Vue.js in the backend out of the box.

### DIFF
--- a/resources/assets/js/backend/before.js
+++ b/resources/assets/js/backend/before.js
@@ -8,3 +8,9 @@
 require('../bootstrap');
 require('pace');
 require('../plugins');
+
+window.Vue = require('vue');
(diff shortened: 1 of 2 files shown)


bambamboole:
I think Vue should not be mounted on the body.

damieneskimo:
@bambamboole Yes You are right. I have added a wrapping div. Can you please check whether it is OK? Thanks

rappasoft:
This is already included in the frontend js, there are no vue components in the backend hence my not loading it.

damieneskimo:
Fair enough. I will close the ticket then.

PR #1119, opened 2018-08-08
fixed social accounts token length

Onatcer:
The current token length in the social_accounts table are unsufficient for the facebook token lengths. I experienced the issue myself, unfortunately one Facebook app review failed because of this. 

There is no fixed maximum length for the facebook access_token, as facebook [employees stated on StackOverflow](https://stackoverflow.com/questions/4408945/what-is-the-length-of-the-access-token-in-facebook-oauth2) but they advised to 255 characters. That's the best i could find. 

Or do you think it should just be ->text() ? 

(Sorry for the double PR, i overlooked the contribution guidelines)


rappasoft:
Interesting, I guess we make it text until Facebook figures out their minds. I'll close this and put it in the next release.

PR #1124, opened 2018-08-12
Using lang directive

rivajunior:
This is a **enhancement**.

I think `@lang('path.to.translation')`  is nice to be used in blade files, instead of `{{ __('path.to.translation') }}` that is less readable and spend more caracteres.

Other think that I've made was the exclusion of every space between directives names and parenteses. Some directives call was with spaces, some not. So I've thinked that is better adopting a pattern for this.


poseso:
{{ __('path.to.translation') }} works for both json translation and normal translation .  @lang does not

rivajunior:
@poseso I've made a local test using both ways `{{ __('string') }}` and `@lang('string')` with a pt_BR.json instead of a directory with php files. Both worked the same way.

The problem that I've found is that in json files I could not use nested keys like 'roles.administrator' because translations in json have a different objective than php. And that occur with both ways.

I think if the person that will use a json file to translations will have to modify every reference to it, no matter the way that the project are doing that.

Useful links:

- https://laravel-news.com/json-based-translations
- https://laravel.com/docs/5.6/localization#retrieving-translation-strings

PR #1141, opened 2018-09-16
Update repository scopes

rawilk:
This PR adds the magic method __call to the BaseRepository to add query scopes easily to repository queries on the fly. It also replaces implode with the spread operator when setting the scopes because implode was causing all parameters to be sent as one parameter, just comma deliminated. 

Adding the scopes through magic methods makes sense to me since it is more like Eloquent this way, which makes it more intuitive to use.

### DIFF
--- a/app/Repositories/BaseRepository.php
+++ b/app/Repositories/BaseRepository.php
@@ -479,6 +479,7 @@ abstract class BaseRepository implements RepositoryContract
     public function __call($scope, $args)
     {
         $this->scopes[$scope] = $args;
+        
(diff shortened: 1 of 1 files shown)


rappasoft:
I've actually been looking for a new base repository but i'll definitely add this for now. Can you just give an example of the __call for anyone that is searching through the issues for info.

rawilk:
An example could be using the `active` scope on the user model like this:

```php
$users = $this->userRepository->active()->orderBy('id')->get();
```

PR #1150, opened 2018-10-01
Add new feature

Nacho1991:
Is this a feature:

- Access language resources through JavaScript. Example lbp.localization.localize("frontend.user.administration","navs")
- Create a token to consume API services after authenticating
- Add the library DataTable
- Add the library Toastr for notifications use


NewEXE:
@Nacho1991, thank you for your's Api part from PR, I used this in my current project :+1:

Nacho1991:
@NewEXE Excellent :-)

lukecurtis93:
just my two cents but passport/datatables/toastr feels a little too opinionated to be included in a boilerplate project, equally im not sure creating a token everytime the user logins and storing that in a session is best practice for this sort of thing

Nacho1991:
@lukecurtis93 you have reason to use (passport / data / toastr) is a matter of preference and I implemented it to facilitate it.

To manage the token there are several ways to implement it, I'm working on that because I also think it's not the best practice :-)

Thanks for the comment :-)

PR #1177, opened 2019-01-18
Enhancement - Show Avatar Image on Edit Front End User

anujjaha:
#enhancement
- Show Font in White Color with Purple Background ( earlier text wasn't visible enough )
- Show Avatar Image in Front User Edit ( earlier it wasn't showing uploaded picture )


rappasoft:
Thanks but this can be added afterwards. I'm trying to cut down on the amount of extra code I need to be responsible for.

PR #1178, opened 2019-01-21
Fixed duplicate input element

anujjaha:
Showing Error In Console ( Duplicate Element with the same element id )
Fixed - by Removing duplicate Id.

### DIFF
--- a/public/css/backend.css
+++ b/public/css/backend.css
@@ -17069,3 +17069,6 @@ body {
   padding-bottom: 1rem;
 }
 
+.white-font {
+    color: white;
+}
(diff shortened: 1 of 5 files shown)


public/css/backend.css
```diff
@@ -17069,3 +17069,6 @@ body {
   padding-bottom: 1rem;
 }
 
+.white-font {
```
JeremyGuinn: Directly modifying the compiled css will result in it being deleted whenever laravel-mix recompiles the public assets.

Also, there already exists a css class to specify white text: `.text-white`

public/css/frontend.css
```diff
@@ -10475,3 +10475,11 @@ a.text-dark:focus {
   max-height: 90px;
 }
 
+.profile-image {
```
JeremyGuinn: Same comment as above, this is directly modifying the compiled css. This will be deleted whenever recompiled with `npm run dev` or `npm run prod`

PR #1187, opened 2019-02-11
setting resource root

SimZal:
Currently resources in CSS get loaded only if you host your site on "domain.xy". This will work even if you have your Laravel on "domain.xy/app/public".

Fix found on: https://www.samundra.com.np/integrating-font-awesome-with-laravel-5.x-using-webpack/1574


rappasoft:
Queued to be merged in the upcoming release.

Issue #1195, opened 2019-02-28
HTML not being rendered in message


vool:
HTML not being rendered in flash messages, eg on the login page when user is not verified, which means the link is not rendered:

![image](https://user-images.githubusercontent.com/441840/53575603-cc7f2400-3b69-11e9-9c85-ed3eaf0f8ebb.png)

Seems to have been introduced with commit fa344c9c488255a6a7589e481d9cb2645b176454, surely there is no risk here as string are not user generated.

Propose reverting fa344c9c488255a6a7589e481d9cb2645b176454 ?

rappasoft:
Reverted: https://github.com/rappasoft/laravel-5-boilerplate/commit/194c2e9e814684b0d4b48a9a7c5db06972f86291

PR #1196, opened 2019-03-08
Fix issue with HTML not being rendered in message

vool:
Re issue #1195

HTML not being rendered in flash messages, eg on the login page when user is not verified, which means the link is not rendered:


rappasoft:
I see this is a security risk, i'm going to try to think of something better than not escaping the output.

rappasoft:
Queued to be merged in the upcoming release.

PR #1215, opened 2019-03-29
composer self-diagnosis & jest

bingalls:
Added `composer self-diagnosis` and `roave/security-advisories`
Updates package.json to pass this test, as well as `npm run test`


rappasoft:
Merged for the other items but the test still does not pass on my end. Also had to update babel-jest because npm audit said it had a security issue.

https://github.com/rappasoft/laravel-5-boilerplate/commit/2fd8d5ddcfcf3e85ce564e4daf2e2c2253c2bc6a

Issue #1216, opened 2019-03-29
5.8 Suggestion : improve package.json compatibility with yarn


bryanrscott:
I would have put in a pull but I already have one open.

Suggestion is to improve package.json compatibility with yarn

In the scripts section of package.json replace npm calls with $npm_execpath
eg
  "dev": "npm run development",
becomes
  "dev": "$npm_execpath run development",

if your using npm it runs unchanged.   if your using yarn in your original call it will utilise yarn instead of npm

rivajunior:
Nice tipe!

I didn't know that `$npm_execpath` variable. I always changed `npm` for `yarn` in my projects

rappasoft:
To my I feel like package.json has a bunch of stuff we don't need. If anyone has the time to look and see if we can thin it at all it would be appreciated.

rappasoft:
@bryanrscott I like the tip but I found this: https://stackoverflow.com/a/53897160/1741608

Might cause an issue for some people if I change it one way or the other.

rivajunior:
> To my I feel like package.json has a bunch of stuff we don't need. If anyone has the time to look and see if we can thin it at all it would be appreciated.

I'll do that.

rivajunior:
> @rappasoft @bryanrscott I like the tip but I found this: https://stackoverflow.com/a/53897160/1741608
> Might cause an issue for some people if I change it one way or the other.

I think that using cross-env it can solve that problem.

I'll try that too.

bryanrscott:
@rappasoft agree it might cause an issue.

PR #1221, opened 2019-03-30
Enhancement build, structure and fix jest tests

rivajunior:
This is both, fix and enhancement.

On this pull request I've made the following changes:

* [Changed](https://github.com/rivajunior/laravel-5-boilerplate/blob/node_optimizations/resources/sass/frontend/app.scss#L5) `_custom.scss` for `_variables.scss`, how [Laravel uses](https://github.com/laravel/laravel/tree/master/resources/sass)

* [Added Fontawesome web](https://github.com/rivajunior/laravel-5-boilerplate/blob/node_optimizations/package.json#L19) on [scss](https://github.com/rivajunior/laravel-5-boilerplate/blob/node_optimizations/resources/sass/frontend/app.scss#L18) to remove Fontawesome SVG used on JS, because it doesn't work on old browsers since Mix isn't processing `node_modules` dependencies. (Plus: build size gets lighter)

* Simplified [.babelrc](https://github.com/rivajunior/laravel-5-boilerplate/blob/node_optimizations/.babelrc) putting [browserlist configuration](https://github.com/rivajunior/laravel-5-boilerplate/blob/node_optimizations/package.json#L43) on `package.json`. In addition to that, I've fixed deprecation warning because of use of `uglify` option who have changed to `forceAllTransforms`.

* [Enforced Unix newlines](https://github.com/rivajunior/laravel-5-boilerplate/blob/node_optimizations/.gitattributes) for some files, like [bootstrap does](https://github.com/twbs/bootstrap/blob/master/.gitattributes)

* [Fixed Jest tests](https://github.com/rivajunior/laravel-5-boilerplate/blob/node_optimizations/package.json#L47).

    * [Added Jest dependencie](https://github.com/rivajunior/laravel-5-boilerplate/blob/node_optimizations/package.json#L27) who was missing.

    * [Changed](https://github.com/rivajunior/laravel-5-boilerplate/blob/node_optimizations/tests/Javascript/Example.spec.js#L1) deprecated `shallow` to `shallowMount`

* Made scripts on [package.json](https://github.com/rivajunior/laravel-5-boilerplate/blob/node_optimizations/package.json#L7) work with _Yarn_ and _NPM_, using `$npm_execpath` how as suggested on #1216.
    * The problem is a Yarn warning which says basicaly that it won't work in the future:

       * > warning From Yarn 1.0 onwards, scripts don't require "--" for options to be forwarded. In a future version, any explicit "--" will be forwarded as-is to the scripts.

* Removed unnecessary `node_modules/.bin/whatever` path from scripts on `package.json`. See [npm-run-scripts](https://docs.npmjs.com/cli/run-script.html)


rappasoft:
Great PR. Thanks for listing all the changes with sources, I have no problem with any of this I just need to test it all my self before merging in.

rivajunior:
> Great PR. Thanks for listing all the changes with sources, I have no problem with any of this I just need to test it all my self before merging in.

I forgot to say that I've tested the scripts on my windows 10 (bash, cmd, powershell) and on ubuntu 18.04 (homestead)

rappasoft:
The only difference I see is that font awesome can no longer be used on the backend, I get that coreui uses simple line icons (which is annoying to begin with) but we should still have the option to use them globally. So i'm probably going to modify that.

rivajunior:
> The only difference I see is that font awesome can no longer be used on the backend, I get that coreui uses simple line icons (which is annoying to begin with) but we should still have the option to use them globally. So i'm probably going to modify that.
> 
> Edit: There are actually a few on the backend that don't currently appear because of this, so we have to make it global somewhere.

Don't worry, I'll do that.

I don't have sure, but I think we don't need to use Simple Line Icons. It is used on CoreUI examples, but we not obligated to use it. So if we remove and use only font-awesome, it can be lighter than use two icon fonts.

rappasoft:
I did it already: https://github.com/rappasoft/laravel-5-boilerplate/commit/7b3c896687f81f92c1c1ad7a0f8fb32ca687d563

PR #1222, opened 2019-03-31
Using PHP-CS-Fixer the right way

rivajunior:

This is a enhancement

The past configuration of [PHP-CS-Fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer) didn't do much and [StyleCI](https://styleci.readme.io/) haved always to do the work.

Now I've looked on StyleCI  [Laravel presets](https://styleci.readme.io/docs/presets#section-laravel) (outdated by the way) and configurate PHP-CS-Fixer by myself.

I must warn you that some configurations are a little opinionated, so I decided to leave the first use of the fixer for you @rappasoft, because a lot of files will be changed and you will be able to decide which configuration to keep.

PS: You can check some examples on [php-cs-fixer-configurator](https://mlocati.github.io/php-cs-fixer-configurator/#)

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -1,0 +1,141 @@
+#!/usr/bin/env php
+<?php
+
+$rules = [
+    'array_indentation' => true,
+    'array_syntax' => ['syntax' => 'short'],
+    'binary_operator_spaces' => ['default' => 'single_space'],
+    'blank_line_after_namespace' => true,
+    'blank_line_after_opening_tag' => true,
+    'blank_line_before_statement' => [
+        'statements' => [
+            'continue', 'declare', 'return', 'throw', 'try', 'do', 'if',
+            'switch'
+        ]
+    ],
+    'braces' => true,
+    'cast_spaces' => ['space' => 'single'],
+    'class_attributes_separation' => true,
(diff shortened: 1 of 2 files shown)


rappasoft:
Just curious what makes StyleCI's current Laravel presets outdated?

rivajunior:
Take it by example:

* **align_phpdoc** should be **align_multiline_comment**

* **blank_line_before_return** is DEPRECATED

    * > An empty line feed should precede a return statement. DEPRECATED: use blank_line_before_statement instead.

From [PHP-CS-Fixer Usage Documentation](https://github.com/FriendsOfPHP/PHP-CS-Fixer#usage)

rivajunior:
@rappasoft have you seen how the code goes with these settings?

If you want I can automatizate it with [Husky](https://github.com/typicode/husky), so everybody who works on the code will normalize the code durring commit, on their own machines.

rappasoft:
At first glance i dont have any issue with what i'm seeing in the 395 files that it modified.

PHPStorm 2019.1 added native support to code cleanup so i'm going to run both and see how they differ.

Screen Shot 2019-03-31 at 22 12 20

PR #1227, opened 2019-04-02
Update AuthServiceProvider

bryanrscott:
# Is this an enhancement
Assign all permissions to the Admin role without the need to explicitly assign  the roles/permissions to the user.

As per spatie recommendation here : https://github.com/spatie/laravel-permission/wiki/Global-%22Admin%22-role


lukecurtis93:
isn't the role called `administrator` on this boilerplate?

bryanrscott:
Updated to use config setting for admin role name

poseso:
you should use Gate; instead of Illuminate\Support\Facades\Gate; else all tests will fail

bryanrscott:
Updated import as suggested

PR #1230, opened 2019-04-07
Update composer.json

akiyamaSM:
Add laravel controllers generator from routes file


Shuyinsama:
We don’t need this at all. Looks like a way of adding his own packages to a boilerplate?

PR #1247, opened 2019-06-19
Update package.json

egdavid:
# Fix
Had to install `webpack-dev-server` and to set `disableHostCheck` to `true` to run `npm run hot` otherwise webpack was throwing the "`Invalid Host/Origin Header`" error and was blocking the live reloading.
This might be useful to those who want to use run hot on Vue.js components development.


rappasoft:
I wonder how much extra this actually adds?

PR #1248, opened 2019-06-23
Update dependencies

rivajunior:
This is an enhancement that updates laravel files to the most recent commit and update node and composer dependencies.


rappasoft:
Sorry I totally missed this and ended up going through the laravel commits my self and updating, but I cherry picked some other ones I missed. Thanks.

PR #1249, opened 2019-06-27
Add disableExceptionHandling function for easier testing in TestCase.php

jason-guru:
### Enhancement
**Brief**: Added `disableExceptionHandling` function to `TestCase.php`
**Explanation**: While testing, some of the test function needs the exception handling to be turned off. And doing it manually leaves a room for error and when code is push to server without undoing the handler change it has cause havoc. So, I have added a function called `disableExceptionHandling` in the `TestCase.php` which just disables the exception handling with a single line of code, where it is needed. 

Also, I have added a test file to test the above function.
Hope, others will find it useful :)

### DIFF
--- a/tests/TestCase.php
+++ b/tests/TestCase.php
@@ -70,20 +70,20 @@ abstract class TestCase extends BaseTestCase
 
     /**
      * Disables the exception handling function and throws the exception instead.
-     * 
+     *
      * @return Throwable $e
      */
     protected function disableExceptionHandling()
     {
-        $this->app->instance(ExceptionHandler::class, new class extends Handler{
-            public function __construct() 
+        $this->app->instance(ExceptionHandler::class, new class extends Handler {
+            public function __construct()
             {
-
             }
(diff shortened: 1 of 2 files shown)


rappasoft:
Can you give a few examples where this is needed in the current test suite? Or is it just not right now but is just a nice to have?

jason-guru:
Hi, thanks for the response. It is not needed in the current test suite, and yes it is a very nice to have feature. For an example when a test assertion like `$response->assertStatus(200);` throws an error `Expected status code 200 but received 400` , and it does not give us a detailed message of what went wrong, but if we use the above function it gives the whole message and traces. I find it really handy to quickly check and rectify my test. :)

rappasoft:
Will come back to this if it's ever needed in the test suite for this project. (Will use it in my personal projects though) as i'm trying to limit unnecessary code.

Thanks.

jason-guru:
Thanks Anthony :) I am glad that you have found it useful.

PR #1258, opened 2019-08-12
NIST Password Rules

DivineOmega:
This PR adds new default validation rules that follow the password related recommendations found in [NIST Special Publication 800-63B section 5](https://pages.nist.gov/800-63-3/sp800-63b.html#sec5). It does so by making use of the [langleyfoxall/laravel-nist-password-rules](https://github.com/langleyfoxall/laravel-nist-password-rules) pacakge.

This supersedes the current direct use of my package ([divineomega/laravel-password-exposed-validation-rule](https://github.com/DivineOmega/laravel-password-exposed-validation-rule)) as the  password exposed validation rule is included in this new package.

I believe this password policy will help new developers make secure choices by default, with the ability to scale complexity up or down as necessary.

Note that relevant tests have been modified, and all pass successfully.


rappasoft:
Do you think we need the current [ChangePassword](https://github.com/rappasoft/laravel-boilerplate/blob/master/app/Rules/Auth/ChangePassword.php) anymore? I feel like it's just added fluff with this new added package. If you think not i'll remove it from all the required places.

rappasoft:
If we're going strictly by NIST, then I vote the ChangePassword rule be removed as NIST says 'No other complexity requirements for memorized secrets SHOULD be imposed.' So i'll go ahead and do that, because they really only add the uppercase/number rules and as of now there's like 10 errors if you set the password as say 'secret'.

rappasoft:
Made some changes and ended up with this: https://github.com/rappasoft/laravel-boilerplate/commit/721d0b5b2aedb390fb45147f3ad0a19e273e4f2d

DivineOmega:
Thanks for getting this all merged in. :+1: :slightly_smiling_face: 

Give me a shout if you have any problems, or file an issue on the package repo.

PR #1285, opened 2019-10-15
Update please

hexpit:
this is a bug.


poseso:
Can you explain why is this a bug?

PR #1299, opened 2019-11-22
Fixed inaccessible dropdown on user management page

### LINKED ISSUE #1298: CSS bug on user management page
Version 6.0.1

There's a bug on the user management page. The actions dropdown on users is hidden behind a weird overflow on the `.table-responsive` element. This is causing the delete action to not be visible. I've found a fix for it that solves the reason for the `overflow-x: auto` property AND the dropdown being inaccessible, that I will push to a PR in a few minutes. Just gotta test it a little more.

Screenshot of the bug happening:
screenshot of bug

mwkcoding:
This fixes #1298

### DIFF
--- a/resources/sass/backend/app.scss
+++ b/resources/sass/backend/app.scss
@@ -6,6 +6,17 @@
 // Import styles
 @import "~@coreui/coreui/scss/coreui";
 
+// Fix inaccessible dropdown in .table-responsive dropdown buttons
+.table-responsive {
+    overflow-x: inherit;
+    td:not(.btn-td) {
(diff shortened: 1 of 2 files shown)


skys215:
I have encountered styling issue due to this commit.

My table's class is set to table-responsive, and the content of the table will overflow at right side. 

After debugging, found out that the `overflow-x: inherit` set in `app.scss` causes this problem. after commenting out this line, the content won't be overflowed. Should I set `overflow-x` to `auto` to show properly? I'm not good at CSS, not sure about it.

PR #1322, opened 2020-02-21
Fix critical issue social login

lloricode:
We just assume that user can change email on their social like facebook,

This fix will able to fetch same data even they update the email of their social, they can fetch same data as they first register here via social


rappasoft:
Wait i'm kind of confused, what's the actual critical bug?

lloricode:
it just search by email,

Issue #1393, opened 2020-06-30
7.0 User Attributes


bryanrscott:
Preparing to migrate our app to the new base template.  Comparing user attributes and have few questions, actually more would you accept pull request(s) along the lines of

a) additional attributes on the user ?
- UUID

b) option to disallow password login (ie only social login) per user?

c) config option to allow only permissions via roles.  The effect would be to not list the individual permissions in the user edit screen.

d) built in end points for extending user attributes.
    eg something like. @includeIf('backend.auth.user.includes.custom-user-attributes')  
         and a call to an empty function applyCustomAttributes($user, array) in userService

rappasoft:
I gave you the UUID trait back (but didn't add it to the user because there's no need right now) because I see that it will probably get used more often than not in a project.

I also gave you the ability to turn off additional permissions: https://github.com/rappasoft/laravel-boilerplate/commit/76a109d116f7078360c34a160e519774c7b03fde

I do not see the need for disallowing password login, just turn off registration. That will force users to use social login, and your admins will still be able to login.

PR #1398, opened 2020-07-04
User options for V7

bryanrscott:
#1393


rappasoft:
See notes in the issue.

PR #1418, opened 2020-07-27
Czech Language translation

xnekv03:
Added Czech language translation


xnekv03:
Added validation strings originaly from [this repo](https://github.com/caouecs/Laravel-lang/tree/master/src/cs)

PR #1428, opened 2020-08-02
Backend : Improve log page integration

bryanrscott:
Mostly ported from v6 forward.   Missing Breadcrumbs and side menu highlighting.


rappasoft:
I'll add the dropdown, but i'm going to leave out the vendor override for now because i'll need to research the breadcrumbs because it's a different package, and some of the font awesome icons aren't rendering in the _master file.

bryanrscott:
Great.

PR #1432, opened 2020-08-06
Update Social Login.   Check if user with email exists and associate.   Also respect the registration config option

bryanrscott:


rappasoft:
If it's via provider ID wouldn't that make the other check not matter?

bryanrscott:
We disable registration and precreate the accounts (so that don't yet have a provider id).    This is to pick up that.

bryanrscott:
Arh.  yes.

rappasoft:
Is this PR still needed?

bryanrscott:
The case is when registration is disabled.    How to create oauth accounts.   Current code always creates a new record.

rappasoft:
If registration is disabled wouldn't you want social turned off as well if they don't already have a social account, seeing as it's technically registration if its the first time?

bryanrscott:
We run with registration disabled and we manually create new users including users who would login socially

rappasoft:
Hmm this seems like a really specific use case. What if a user registers with an email, then tries to login socially with the same email? Shouldn't that not be allowed? If you do, then they can now log in socially and with a password for the same account. Also i'm confused with where the registration check is in your code, shouldn't it be at the top of the method?

bryanrscott:
Maybe we are a specific case.   We register all our users first.
We do add an attribute password logon to each user, which is generally disabled for oauth users.

PR #1450, opened 2020-09-11
Missing name attribute added to UserTable

mkysoft:
name attribute not showed because of missing attribute parameter.

### DIFF
--- a/app/Http/Livewire/RolesTable.php
+++ b/app/Http/Livewire/RolesTable.php
@@ -32,10 +32,10 @@ class RolesTable extends TableComponent
     public function columns(): array
     {
         return [
-            Column::make(__('Type'))
+            Column::make(__('Type'), 'type')
                 ->view('backend.auth.role.includes.type', 'role')
                 ->sortable(),
-            Column::make(__('Name'))
+            Column::make(__('Name'), 'name')
                 ->searchable()
                 ->sortable(),
             Column::make(__('Permissions'), 'permissions_label')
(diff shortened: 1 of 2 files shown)


dregad:
@mkysoft thanks, I confirm this fixes the problem. 

The same fix should be applied to the `Type` column too, as well as in the RolesTable class for both columns.

rappasoft:
What issue is this causing you? This is not needed, it is inferred: https://github.com/rappasoft/laravel-livewire-tables/blob/master/src/Views/Column.php#L79

mkysoft:
> What issue is this causing you? This is not needed, it is inferred: https://github.com/rappasoft/laravel-livewire-tables/blob/master/src/Views/Column.php#L79

You are right now, code use label as field name. When I created this pull request firt time, Name field not displaying on User table. I made this changes because of @dregad comment.

dregad:
> What issue is this causing you? This is not needed, it is inferred: https://github.com/rappasoft/laravel-livewire-tables/blob/master/src/Views/Column.php#L79

@rappasoft I did not re-test this against the latest development branch, but the problem is that you're passing the column's *label* to the constructor, so the logic only works as long as the lowercase label matches the column's name, which is often not the case in localized settings. See #1457 for details.

rappasoft:
Ah ok I understand now.

rappasoft:
Adding this manually because the massive refactor I did makes it easier that way right now. Thanks for finding the issue though!

PR #1486, opened 2020-12-18
add GOOGLE_ANALYTICS to .env.example

taoufiqaitali:


rappasoft:
Ended up removing GA in the next release. Easy to add yourselves as GA is more complicated than it was back when that code was added.

PR #1490, opened 2020-12-27
Update UserService.php

Azmeer-crescent:
Socialite login (eg: google) doesn’t handle existing accounts, so I made this change and it worked.


bryanrscott:
Do you think you should update the provider_id for existing accounts discovered?

Issue #1589, opened 2022-07-08
Deactivated user can login by requesting password reset


redactuk:
Steps on clean laravel-boileplate v8.3.1 install:

1. Admin deactivates a user
2. Deactivated user can't login now, but can click 'Forgot your password?' link
3. Deactivated user then resets password using email link sent, and after changing password is automatically logged in

Pretty sure this is not meant to happen, as if this user then logs out again, once again they can't login as 'deactivated'

bryanrscott:
Yes you are right.

A possible solution would be to modify the rules function in the file app\Domains\Auth\Http\Controllers\Frontend\ResetPasswordController.php 
to update the 'email' requirement.

            'email' => ['required', 'max:255', 'email',
                Rule::exists(with(new User())->getTable())->where(function ($query) {
                    return $query->where('active', '=', true)->whereNotNull('email_verified_at');
                }), ],

Though this would still send the email with token to the user.

redactuk:
What I was trying to work out is where user is automatically logged in after password reset, as surely there would be the point at which a check is done to see if the user is desctivated? i.e. let them change password, but still block them logging in once that done.

bryanrscott:
I think that is in https://github.com/laravel/ui/blob/3.x/auth-backend/ResetsPasswords.php

im-denny:
Hello,

I have submitted pull request #1596 addressing this issue. 

I wrote few lines of code intervening redirection after successful reset password attempt, which will check for user's active status beforehand proceeding redirection and log them out incase found inactive. I have written test for the same as well.

ccsliinc:
@im-denny I have made a new pull request to fix this issue. #1599

PR #1599, opened 2023-01-26
Fixes inactive user logging in via reset password link.

ccsliinc:
If a user has an inactive account and tries to reset the password it will redirect them to the login page with an error. The notification will not be sent.


ccsliinc:
This resolves #1589

